PR #10, opened 2015-02-22
Use file_name_width in grid

coyotebush:
Filenames with characters that need multiple bytes in UTF-8, and/or
control characters like combining diacritics, would break the grid.
`StrExt::width` seems to do the right thing, and there's conveniently a
function here (which was otherwise unused) to call that.


ogham:
Good catch, thanks!

PR #11, opened 2015-02-22
Sorting improvements

coyotebush:
Several proposed fixes to sorting by extension, size, or date, for better consistency with itself and GNU `ls`. Feel free to disagree with any part of this.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -5,7 +5,6 @@ use column::Column::*;
 use output::{Grid, Details};
 use term::dimensions;
 
-use std::ascii::AsciiExt;
 use std::cmp::Ordering;
 use std::fmt;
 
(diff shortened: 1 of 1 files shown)


ogham:
I don't disagree, I'm just not sure - does it make more sense to have the largest files (or the most-recently-modified files) be on the top or bottom by default? I went with having them sort ascending instead of descending, as usually I use `--sort=size` in order to find the largest file in a directory, and if there are many files, then the ones I want to see get scrolled off the screen.

coyotebush:
That's a good point. My fingers are used to `ls -lt | head` for that pattern, but I'd be happy to adapt (and put that commit aside for now) if you want to make the decision to differ from `ls` here.

ogham:
Yeah, I think that's the decision I'd like to make. Having the heaviest files 'fall to the bottom' by default makes the most sense to me.

I completely agree with the changes you've made to sorting by extensions, though!

ogham:
Cherrypicked the extension changes in ee20c5d8bb1fdf0eae381a76218aae2e2ae78546.

coyotebush:
Oops, I was simultaneously removing the ordering change from the branch. Just for future reference, it was https://github.com/coyotebush/exa/commit/cf88b93c4cf07361b0f17855d6b13f6ff9d68e00

ogham:
The problems of rewriting history :)

PR #14, opened 2015-02-22
Add support for extended attributes

nwin:
exa now matches the behavior of `ls` on OSX. In theory this could also work on Linux but the API is different and I can’t test it. It should be relatively easy to implement it for Linux as the rust API is already generalized.

I haven't tested it unter linux but it should work since the feature is a no-op there.


nwin:
I’ve no clue why travis failed, maybe rustc is too old.

ogham:
I'm with Travis on this one - you need to bring the xattr module into scope before you can use it in tests. Try adding `use xattr;` into the block of `use`s on line 476, then test again.

nwin:
Of course… :(

ogham:
>> #25

PR #17, opened 2015-02-22
Fix the column alignment

nwin:
Count the number of grapheme clusters to determine the cell length.

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -86,7 +86,7 @@ impl Cell {
     pub fn paint(style: Style, string: &str) -> Cell {
         Cell {
             text: style.paint(string).to_string(),
-            length: string.graphemes(true).count(),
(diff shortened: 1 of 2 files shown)


nwin:
Ah, my bad. The joy of boolean arguments…I assumed it would mean the same as in StrExt::graphemes.

ogham:
* length → width: Good call, `len` is completely the wrong thing to use here!
* extra year padding: you know, I was looking at a mixed year-and-time directory yesterday, thinking it didn't look quite right, but I couldn't put my finger on why. Adding that extra space makes it look *much* better.

Thanks!

nwin:
For optimizations one could think of using width only for the username, group, date and the filename. The other fields should never contain non-ascii chars.

PR #19, opened 2015-02-22
Get rid of explicit `as_slice()` calls.

Stebalien:
This is obviously completely unnecessary but `as_slice` will be going away eventually.


ogham:
Can't say fairer than that! Thanks.

PR #21, opened 2015-02-22
Improve matching of Git status entries to files

### LINKED ISSUE #15: Does Not Show Git Status in Subdirectories
If we have a modified file in a subdirectory (related to git root folder), the subdirectory is marked as modified, but the file is not. Please see the screenshot

https://www.dropbox.com/s/n2zmyacreuowb78/Screenshot%202015-02-22%2015.40.41.png?dl=0

coyotebush:
The challenge is that the paths returned from libgit2's status listing
are from the perspective of the Git repository and thus effectively
relative to the working tree root, while the other paths we're
manipulating are (potentially) relative to our current working
directory. So, if those two aren't identical (if running from outside
the working tree, or from a subdirectory), the paths won't match up.

A reasonably reliable way around this is to resolve both types of paths
to absolute paths before comparing them. This fixes #15 at a basic
level, anyway.

What still doesn't work: referring to the working tree or one of its
descendants via a symlink. For that, we'd probably need to fully resolve
symlinks in the file path.

(The unwrap_or()'s are messy and will probably just result in missing
status information, but then, what information could you hope to get
without having both a current working directory and a Git working tree?)

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -75,7 +75,7 @@ impl Dir {
 /// Container of Git statuses for all the files in this folder's Git repository.
 #[cfg(feature="git")]
 struct Git {
-    statuses: Vec<(Vec<u8>, git2::Status)>,
+    statuses: Vec<(Path, git2::Status)>,
 }
 
 #[cfg(feature="git")]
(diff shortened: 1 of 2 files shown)


ogham:
This is another of those things that I could have swore I actually tested, but didn't... thanks again!

PR #23, opened 2015-02-23
Optimize Builds

Stebalien:


ogham:
Good idea! I didn't know about `lto`.

petevine:
I didn't about it being on by default - it conflicts with `-C codegen-units` which I started using recently, hence the discovery :) (build failure)
This is one of the reasons it's not being turned on by any of the opt levels.

PR #25, opened 2015-02-23
Add xattr for linux

nwin:
I made it as a separate PR since I’m not sure if this is the way to go or if we need to introduce a feature since linux might not have extended attributes built into the kernel but I think this is common.


ogham:
Does this completely supersede #14?

nwin:
Yes this branch is based on the other one.

ogham:
I tested this on OS X and Linux - all good.

ogham:
I changed the pre-calculated width of the permissions column from 10 to 11 to account for the new @ character.

PR #31, opened 2015-02-23
Add a man page entry for exa

davidcelis:
Hi! First off, thanks for writing this utility! I've been playing around with it and the git integration is really awesome. I hope this PR isn't too out of the blue... But I (and probably other UNIX users) have a reflex to type `man <command>` to find out more about how to use it. I typically immediately try a `--help` option if that doesn't work, and I'm glad exa has one. But then I thought that it could be pretty neat for `exa` to have a manpage as well. This patch adds a manual page that can be parsed and outputted by the `man` command. Some notes:

 * This manual page was hand crafted. I had hoped to find some Rust
   utility that could generate manual pages to prevent having to change
   this file whenever changes to options are made. There are tools to
   write manual pages in a nicer syntax and then convert them into the
   groff syntax, but I hesitated to add more tools to the build process.
   This might make the file more difficult to maintain, but unless exa's
   signature changes frequently, it should be okay.
 * The manual page isn't automatically copied into the
   /usr/local/share/man/man1 directory, where it would need to be for
   `man` to read it. My first inkling to have installation be a nice
   process would be to craft a Makefile that will run `cargo build
   --release`, copy the binary into /usr/local/bin, and copy the man
   page into the aforementioned directory. If it's alright with you (and
   regardless of whether or not you like the idea of a man page), I'd
   love to handle that in this PR or another one if you like the idea
   of a Makefile to ease installation!
 * I didn't directly and copy the verbiage from the `--help` option; I
   instead chose to carefully reword some things to better match the
   style that man pages (in my experience) tend to have. I hope that's
   okay as well.

Signed-off-by: David Celis <me@davidcel.is>

### DIFF
--- a/man/exa.1
+++ b/man/exa.1
@@ -1,0 +1,112 @@
+.TH EXA "1" "February 2015" "exa 0.1.0" "User Commands"
+.SH NAME
+exa \- A replacement for ls with git integration
+.SH SYNOPSIS
+.B exa
+[\fIOPTIONS\fR] [\fIFILES...\fR]
+
+.SH DESCRIPTION
+exa is a replacement for ls, written in Rust. If built with libgit2, it includes git integration and can display relevant information when used in git repositories.
+
+.SH OPTIONS
+
+.TP
+\fB\-?\fR, \fB\-\-help\fR
+Display a help message
+
+.TP
+\fB\-@\fR, \fB\-\-extended\fR
+Display extended attribute keys and sizes in long (-l) output.
+
+.TP
+\fB\-1\fR, \fB\-\-oneline\fR
+Display one entry per line.
+
+.TP
+\fB\-a\fR, \fB\-\-all\fR
+Display entries whose names begin with a dot (.).
+
+.TP
+\fB\-b\fR, \fB\-\-binary\fR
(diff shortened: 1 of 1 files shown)


crdx:
nice

ogham:
Hi David, thanks for writing the man page - I definitely want to use it. I know what you mean about the language being slightly different for manual pages than `--help` text, and yeah, groff syntax isn't the nicest but at least this one's just one page of text!

However, I'm not sure what exa's installation story is going to be. Right now, the [instructions on the web page](http://bsago.me/exa/#installation) tell the user to git clone and set up an alias in their shell, which is *really* not good. There's also [rustle](https://github.com/brson/rustle), which provides easy installation for non-Rust users, but I don't believe there's any way for it to run an installation script or a Makefile. And maybe eventually Cargo will provide [its own program-installation mechanism](https://github.com/rust-lang/cargo/issues/1334), which, given its ubiquity, would quickly become the default.

I know that a lot of people are used to installing software with `make install`, so if you send in a PR with a Makefile, it'll be gracefully accepted! But, annoyingly, I can't promise that it'll be the default way to install it forever, so please bear that in mind before you start coding.

PR #34, opened 2015-02-24
Fix a missing '/' in symlink targets

### LINKED ISSUE #33: Missing '/' character in symlink target filename
I used exa to list my home directory, and noticed that some of the filenames are missing at least one `/` character.

Specifically, when a symlink should be printed like the following:

```
symlink => dirA/dirB/dirC/file
```

It is actually printed as:

```
symlink => dirAdirBdirC/file
```

jgilchrist:
In cases where symlink targets were more than a single directory down,
exa did not print the `/` characters to separate directories, resulting
in the following output:

    symlink => dirAdirBdirC/file

Instead of

    symlink => dirA/dirB/dirC/file

By adding a `/` character after each component of the filename, this
error is fixed.

Fixes #33

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -162,13 +162,8 @@ impl<'a> File<'a> {
                         for component in path_bytes.init().iter() {
                             let string = String::from_utf8_lossy(component).to_string();
                             path_prefix.push_str(&string);
+                            path_prefix.push_str("/");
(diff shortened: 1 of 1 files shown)


ogham:
argh I can't believe I did that. Thank you!

PR #35, opened 2015-02-24
Add simple Makefile to build and install exa

davidcelis:
This patch adds a Makefile that will build and install exa along with
its recently added manual page. This provides a little bit of a nicer
installation path while we wait for Cargo to, hopefully, provide its own
built-in installation method for packages that contain binaries.

Signed-off-by: David Celis <me@davidcel.is>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+PREFIX ?= /usr/local
+
+BUILD = target/release/exa
+
+$(BUILD):
+	@which rustc > /dev/null || { echo "Exa requires Rust-lang to compile. For installation instructions, please visit http://rust-lang.org/"; exit 1; }
+	cargo build --release
+
+build: $(BUILD)
+
+INSTALL = $(PREFIX)/bin/exa
+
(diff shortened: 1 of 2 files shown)


ogham:
Thanks again! I'll update the website to reflect the new instructions.

ogham:
Ohwait. Hang on. Is it possible to edit the Makefile to allow it to pass `--no-default-features` to Cargo somehow? I feel your Makefile-fu is better than mine.

davidcelis:
So I actually started the Makefile with a check on the presence of `libgit2`. I'm not sure how to assume where it's installed (mine is `/usr/local/lib/`) so I figured I'd try building it and, if the build failed, attempt to re-build with `--no-default-features`. But then I was surprised to see that `exa` built successfully without `libgit2` installed and, from what I could tell, displayed git information fine. I don't think OS X comes with a version of `libgit2`, so... I'm not really sure what happened there!

ogham:
That's weird! I have no idea either.

PR #45, opened 2015-03-04
making Exa.load() stat files in parallel

byteprelude:
Makes `load()` stat the files in parallel rather than sequentially. This increases performance slightly on my machine.


ogham:
Crazy awesome! :+1:

Issue #47, opened 2015-03-10
Option to omit git status column (or disable by default)


killercup:
As mentioned in #28, exa ist rather slow in larger git repositories. I know there is the possibility to use Cargo's `features` to compile without git support, but that is not what I want.

When using `--long`, the _Git_ columns is shown automatically. I think only enabling it with a `--git` flag might be better.

Alternatively, I'd like to see an option to disable this column.

PR #48, opened 2015-03-10
Make Git Column Optional

killercup:
You need to add the `--git` option if you want to show that git column. It only works if exa was compiled with git as a Cargo feature (default) and the current directory is a git repository.

This is what I described in #47. It is a breaking change for some users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 - **-B**, **--bytes**: list file sizes in bytes, without prefixes
 - **-d**, **--list-dirs**: list directories as regular files
 - **-g**, **--group**: show group as well as user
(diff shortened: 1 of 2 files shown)


killercup:
This was easier than expected. Did I miss something?

ogham:
Thanks for this. It's annoying to have to do this, but I haven't found a solution to why Git viewing is so slow, and it's slowing people down.

> Did I miss something?

Doesn't look like it! I did go through a bunch of different ways to parse the options and glad the final code is easy enough to understand :)

killercup:
Thanks for merging :)

ogham:
:+1:

PR #49, opened 2015-03-10
Add "`--group` is useless without `--long`"

killercup:
Just noticed that some flags printed that error, but `--group` was silently ignored.


ogham:
Oops! :+1:

PR #51, opened 2015-03-22
Two minor tweaks: one to make exa build, one for tests to build

e12e:
Hi,

I just tried building/testing exa locally, with rust nightly. Apparently osStrExt has moved(?). Also not sure if tests were run prior to previous commit -- or if something has happened wrt visibility, but cargo/rust complained that path_filename wasn't public, but private. With the tweak in the tests, the test build and pass.

It's quite possible that you'd want to reject this pr and make (better?) fixes yourself. Just thought it might be of interest in case others run into the same issues.

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -84,7 +84,7 @@ impl Git {
     /// Discover a Git repository on or above this directory, scanning it for
     /// the files' statuses if one is found.
     fn scan(path: &Path) -> Result<Git, git2::Error> {
-        use std::os::unix::OsStrExt;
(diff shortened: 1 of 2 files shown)


ogham:
Hey, thanks for this. I didn't know that `OsStrExt` had moved - I thought that a dependency had broken, and I was waiting for it to be fixed - didn't know that it wouldn't ever be fixed! As for the tests, I'm pretty sure I ran them, so I have no idea if something *did* happen with visibility or something. But whatever happened, your PR makes it work, so thanks.

PR #57, opened 2015-04-09
updated to latest rust nightly

tsurai:
This pull requests updates some out dated dependencies causing compile errors and removes the usage of the AdditiveIterator Trait.

Even though the documentation says otherwise the AdditiveIterator has been removed (https://github.com/rust-lang/rust/pull/23293/) which caused an compile error.

I'm using the newest upstream version directly from github. Travis might be a little behind and fail.


ogham:
Thanks! :+1:

PR #68, opened 2015-06-17
Fix compile for DragonflyBSD and nightly rust

mneumann:


ogham:
Thanks again! :+1: Glad to see some of these features stabilising.

PR #79, opened 2015-09-13
This makes exa work on other platforms than linux and macosx.

mneumann:


ogham:
This is a repeat of #68, isn't it? Sorry for having to make you make this change again :sob:

mneumann:
No problem ;-)

PR #81, opened 2015-09-28
Fix integer length error on 32bit environment

rhysd:
This PR fixes below type mismatch error in 32bit Environment.

```
src/output/details.rs:585:49: 585:60 error: mismatched types:
 expected `i64`,
    found `i32`
(expected i64,
    found i32) [E0308]
src/output/details.rs:585         let date = self.tz.at(LocalDateTime::at(timestamp.0));
                                                                          ^~~~~~~~~~~
src/output/details.rs:585:49: 585:60 help: run `rustc --explain E0308` to see a detailed explanation
error: aborting due to previous error
Could not compile `exa`.
```

My environment:
- Rust nightly 7bf4c885f
- Ubuntu Linux 32bit

### DIFF
--- a/src/output/details.rs
+++ b/src/output/details.rs
@@ -582,7 +582,7 @@ impl<U> Table<U> where U: Users {
     }
 
     fn render_time(&self, timestamp: f::Time) -> Cell {
-        let date = self.tz.at(LocalDateTime::at(timestamp.0));
+        let date = self.tz.at(LocalDateTime::at(timestamp.0 as i64));
 
         let format = if date.year() == self.current_year {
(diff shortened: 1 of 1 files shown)


ogham:
Thanks a bunch! :+1: 

I must ask, though: both you and @petevine found this with a few days of each other, but the code in question hadn't been touched in a while. So can you tell me whether this used to compile fine and then suddenly broke one day? I had a quick look in libstd and couldn't seem to find a change that would cause it to break, which confuses me.

rhysd:
> So can you tell me whether this used to compile fine and then suddenly broke one day?

This is a first build of exa for me.  So I don't know what this bug is derived from.

ogham:
Hm. Weird!

Issue #84, opened 2015-10-01
32-bit x86 Linux build fails


petevine:
Trying to build the latest exa master with rust-1.3.0 on 32-bit Slackware ends with:

````
Compiling scoped_threadpool v0.1.6
Build failed, waiting for other jobs to finish...
failed to run custom build command for `scoped_threadpool v0.1.6`
Process didn't exit successfully: `/home/petevine/exa-master/target/release/build/scoped_threadpool-3092426cc5b469bd/build-script-build` (exit code: 101)
--- stderr
thread '<main>' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:363
````

The rustc used generates pentium2 code by default. (my own build)

ogham:
Hey,

This one isn't failing because you're on a 32-bit install, it's because you're on Rust 1.3.0 (stable). exa requires the nightly because it uses features that have only recently been added, and need some time to stabilise.

The reason that your build is failing on `scoped_threadpool` rather than exa is that the [build script for that crate](bs) checks the version at compile time. There's a bug that's been fixed in 1.4.0 but still exists in 1.3.0 that makes scoped threads unsound but still compile successfully, so it checks the version manually rather than just relying on a feature flag.

I had a warning in the Readme that said you should use the nightly, but that warning is so old that the version numbers are out of date! Now fixed -- sorry if that was the source of your confusion. I periodically go through all the unstable features I'm using in #64. Right now we're just waiting for features to settle down.

I recommend installing the nightly or using multirust if you also want to use stable (multirust is great)

[bs]: https://github.com/Kimundi/scoped-threadpool-rs/blob/9c2e4ab8b5b28bbffe085478694c130f9cae3f5a/build.rs

petevine:
Thanks, I thought I was recent enough.

petevine:
So I repeated the whole build from scratch using 1.5.0 from 20151002 and got the same result - where's the catch? Do I have to clean cargo caches too?

ogham:
Hmm, maybe. It's possible that it's cached the failure from when you tried to compile it last time (I'm not too well-versed on which things stick around and which don't). Try a `cargo clean`?

petevine:
Even though `cargo clean` didn't help, I managed to build exa using a different CARGO_HOME.

The only surprise comes from the size difference between the stripped arm and x86  binaries - like a 1:2 ratio.
Both were built with --release so why the difference?

ogham:
[...]

PR #85, opened 2015-11-04
Move to stable

skade:
This introduces multiple changes to make exa work on Rust 1.4.0. Detailed reasoning behind all patches:

* 679b9e1: The one instance of `sum` can be trivially replaced by `fold(0, Add::add)`, dropping a convenience feature flag.
* b3e3825: `fs_mode`s [tracking issue](https://github.com/rust-lang/rust/issues/27712) currently gives no hope for `fs_mode` to remain as it is or to be stabilized anytime soon. As the only thing imported is a constant table, this can be copied over into the library - it will most likely break in the future anyways.
* e8ea96e: The RFC for CString convenience conversions has failed and the corresponding methods are flagged for [deprecation](https://github.com/rust-lang/rust/issues/27704). Take the route over `String` instead, which avoids many of the issues described in the corresponding [RFC](https://github.com/rust-lang/rfcs/pull/1255) discussion.
* 7f53da7: As the "last" element of the Slice is dropped anyways, this can be expressed using the slicing syntax just as well, dropping the need for a feature that only becomes available in Rust 1.5. Add tests to ensure that the API still holds.
* 7a97b7d: I'm a bit unsure here: I assume that the vector only grows, which allows to use `reserve` instead of `resize`, to my reading. Resize will become available in Rust 1.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,8 @@ before_install:
   - sudo apt-get install cmake
 sudo: true
 language: rust
-rust: nightly
+rust:
+  - nightly
+  - beta
+  - stable
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [exa](http://bsago.me/exa) is a replacement for `ls` written in Rust.
 
-Works on all recent Rust versions > 1.4.0.
+Works on all recent Rust versions >= 1.4.0.
 
 ## Screenshot
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,6 @@ You can sort by **name**, **size**, **ext**, **inode**, **modified**, **created*
 
 ## Installation
 
-exa is written in [Rust](http://www.rust-lang.org). You'll have to use the nightly -- I try to keep it up to date with the latest version when possible.  Once you have it set up, a simple `make install` will compile exa and install it into `/usr/local/bin`.
+exa is written in [Rust](http://www.rust-lang.org). Once you have it set up, a simple `make install` will compile exa and install it into `/usr/local/bin`.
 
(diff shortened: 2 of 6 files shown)


skade:
Yes. For some reason, 1.3.0 throws this:

```
   Compiling exa v0.4.0 (file:///Users/skade/Code/rust/exa)
src/output/details.rs:234:17: 280:19 error: invocation of unsafe method requires unsafe function or block [E0133]
src/output/details.rs:234                 scoped.execute(move || {
src/output/details.rs:235                     let mut errors = Vec::new();
src/output/details.rs:236
src/output/details.rs:237                     let mut xattrs = Vec::new();
src/output/details.rs:238                     match file.path.attributes() {
src/output/details.rs:239                         Ok(xs) => {
                          ...
note: in expansion of for loop expansion
src/output/details.rs:229:13: 281:14 note: expansion site
note: in expansion of closure expansion
src/output/details.rs:225:21: 282:10 note: expansion site
src/output/details.rs:234:17: 280:19 help: run `rustc --explain E0133` to see a detailed explanation
error: aborting due to previous error
Could not compile `exa`.
```

ogham:
1.3.0 fails because the scoped_threadpool crate makes that particular method unsafe for that version to work around a bug in rustc that’s since been fixed. More details here: https://github.com/ogham/exa/issues/84#issuecomment-144882186

skade:
I fixed the tree output by reimplementing resize in terms of a reallocation and a `extend`.

It's certainly not efficient, but it works. I added a TODO to keep track that this should be cleaned up after the 1.5.0 release.

skade:
Moved the constants to a private module.

ogham:
LGTM! :+1:

I’m OK with copy-and-pasting code from libstd in lieu of waiting for some of the features to stabilise, especially if these features are *already* going to be stabilised soon. The benefit of having everyone able to compile exa themselves outweighs the small cost of ever fixing it if we need to. So thanks very much for this.

PR #87, opened 2015-12-11
Fix compilation on arm using stable rust 1.5

petevine:
1.5 is still compatible with the `users` crate so, unlike 1.6, it doesn't need changing anything else to compile.


ogham:
Thanks again! :+1:

PR #89, opened 2015-12-14
Minor fixes for FreeBSD

jbeich:
ioctl() command as integer was also tested on NetBSD and OpenBSD under qemu.
```c
#include <sys/ttycom.h>
#include <stdint.h>
#include <stdio.h>

int main() {
  printf("%#jx\n", (intmax_t)TIOCGWINSZ);
}
```
```sh
$ ./a.out
0x40087468
```

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -727,9 +727,8 @@ LONG VIEW OPTIONS
   -U, --created      display timestamp of creation for a file
 "##;
 
-static GIT_HELP:      &'static str = r##"  -@, --extended     display extended attribute keys and sizes"##;
-static EXTENDED_HELP: &'static str = r##"  --git              show git status for files"##;
-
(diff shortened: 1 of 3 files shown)


src/options.rs
```diff
@@ -727,9 +727,8 @@ LONG VIEW OPTIONS
   -U, --created      display timestamp of creation for a file
 "##;
 
-static GIT_HELP:      &'static str = r##"  -@, --extended     display extended attribute keys and sizes"##;
-static EXTENDED_HELP: &'static str = r##"  --git              show git status for files"##;
-
```
ogham: Oh wow, I can't believe I did this!

ogham:
Thank you for this! :+1:

I wouldn't have known what the ioctl number was for those OSes, so seeing your test case helps a lot.

PR #93, opened 2016-01-07
Document using `cargo install` in README

tomassedovic:
Cargo now lets people install binaries by running `cargo install`. This
can be more convenient than cloning and building exa manually.


tomassedovic:
Hm, the build failure seems to be a broken dependency rather then a problem with this pull request.

ogham:
Thank you for adding this! :+1: 

Also I wonder how you could have broken the build by editing the README...

PR #103, opened 2016-03-13
Update Cargo dependencies

petevine:
Strictly required are the following changes:
- `datetime/locale` to fix ARM (which seem to have never landed)
- `libgit2-sys` to fix nightly

The rest were just routinely updated.


petevine:
I wonder about those windows dependencies being removed - is that right?

ogham:
Thanks, I didn't notice that the dependencies had stayed the same for so long. :+1: 

It looks like the crates that were depending on the `{advapi,kernel,ws2_}32-sys` crates (rand, gcc, libssh, winapi, among others) now just depend on libc, so the dependency graph is a lot less complex.

PR #111, opened 2016-05-02
Fix building without default features

hoodie:
this makes it possible to build with `--no-default-features` again


ogham:
Oh crap, I totally missed this. Thank you again! 👍

PR #115, opened 2016-07-04
Improvements to the makefile

maandree:
See commit messages for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,21 @@ BUILD = target/release/exa
 
 $(BUILD):
 	@which rustc > /dev/null || { echo "exa requires Rust to compile. For installation instructions, please visit http://rust-lang.org/"; exit 1; }
-	cargo build --release
+	if test -n "$$(echo "$$CC" | cut -d \  -f 1)"; then \
+	    env CC="$$(echo "$$CC" | cut -d \  -f 1)" cargo build --release; \
+	else\
+	    env -u CC cargo build --release; \
+	fi
 
 build: $(BUILD)
 
 build-no-git:
 	@which rustc > /dev/null || { echo "exa requires Rust to compile. For installation instructions, please visit http://rust-lang.org/"; exit 1; }
-	cargo build --release --no-default-features
+	if test -n "$$(echo "$$CC" | cut -d \  -f 1)"; then \
+	    env CC="$$(echo "$$CC" | cut -d \  -f 1)" cargo build --release --no-default-features; \
+	else\
+	    env -u CC cargo build --release --no-default-features; \
+	fi
 
 INSTALL = $(PREFIX)/bin/exa
 
--- a/Makefile
+++ b/Makefile
@@ -30,4 +30,7 @@ uninstall:
 	-rm    -- "$(DESTDIR)$(PREFIX)/bin/exa"
 	-rmdir -- "$(DESTDIR)$(PREFIX)/bin"
 
-.PHONY: install uninstall
+clean:
+	-rm -rf target
+
+.PHONY: install uninstall clean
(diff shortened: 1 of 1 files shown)


ogham:
Thanks for this! I never really learned how to Make properly.

ogham:
Ack, I thought I merged this :fearful:

About the most recent commit: is there a reason why listing each Rust file in the Makefile is preferred? (I can see myself forgetting to update the list)

maandree:
The idea is that you should compile each source file as independently as possible to save compilation time. But I haven't look at how to compile rust files so I did not add that. If it was C however (which is particularily well suited for this), this with be done like so (assuming flat file hierarchy however) (you could also nice tricks with the help of the compiler make so that files are not recompiled with .h files that do not affect them are modified)

```
SRC = a.c b.c c.c
OBJ = $(SRC:.c=.o)
my-executable: $(OBJ)
  $(CC) -o $@ $^
%.o: %.c *.h
  $(CC) -c -o $@ $<
```

A second reason, that is mostly ignored today, is so you can have a `dist` rule that create a tarball with all files for when you release a new version. As I wrote in the commit message, you can just use `SRC = src/*.rs src/*/*.rs src/*/*/*.rs` if you only want to compile all files at once and not add any other fancy rules like `dist`. If you also fine with just GNU make support, and not other make implementations, you can even use `SRC = $(shell find src -name '*.rs' -print)`.

kballard:
There's not really any point to listing all the files separately here. The only benefit is having `make` do absolutely nothing if no files have changed. But `cargo` already handles that for you, so really we should just tweak the Makefile to mark `target/release/exa` as a phony target, so that way `make` will always call into `cargo` (and then `cargo` can just choose to do nothing if no files have been changed).

kballard:
I just submitted #173 to make `make` always call through into `cargo`, which should remove the need to list source files in the `Makefile`.

ogham:
Like I said ages ago, I never really _got_ Make. I can usually read Makefiles, I just can’t write them.

There are two things I’m stuck on at the moment with regards to _this_ Makefile. Could anybody help me out?

1) Is there anything that the Makefile will need to have that can’t be done by Cargo? I’m guessing that things like installing man pages or shell completions are out of range for `cargo install`, even though OSes’ package managers will do this.
2) Similarly, is there anything *currently* being done by the Makefile that could be done by Cargo instead? It uses Cargo to build, but not install, and it *looks* like Cargo will only install to a user’s local directory, rather than globally.

I added two more files yesterday, so it looks like not needing the hard-coded list of files is already a plus.

kballard:
`cargo` can do nearly everything, except:

1. It can't install manpages. It only installs binaries.
2. By default it will install into `~/.cargo/bin`. If you want to install into `/usr/local/bin` you need to pass `--root /usr/local` to `cargo`.

If it weren't for the manpages, I'd say remove the `Makefile` entirely and change the installation instructions to `cargo install --root /usr/local`. But since you do have manpages, having a Makefile seems like a reasonable thing to do.

Some of the commits in this PR seem useful, others not so much (who has an `uninstall` rule in a Makefile? And why is it trying to delete `$(PREFIX)/share/man/man1`?)

ogham:
I ended up merging the PRs. It now uses Cargo for compilation rather than listing a bunch of files. I’ve thought about splitting up the source into several crates so we can take advantage of partial compilation, but at the current rate of change, Cargo will land that before exa does.

I also added some rules about installing the man page and some shell completions, though I’m not 100% sure that I’ve got the directories right. And I have seen `uninstall` in the Makefile before, even if it’s just a guideline for what a package manager should be doing. Though I haven’t seen it try to delete the `man1` directory. Preferably there’d be a way to only delete it if it’s empty, but I couldn’t find a cross-platform flag to `rmdir` for that case.

kballard:
Well, `rmdir` only removes a directory if it's already empty, but that's not the point. Having it there will spit out spurious errors. In addition, if the directory existed (but was empty) prior to installation, it will end up deleting the directory even though it didn't create it in the first place.

kballard:
There were still a bunch of issues with the Makefile. I just submitted #176 to fix them.

Issue #76, opened 2015-08-03
panics when outputs to a pipe that closes early


lilydjwg:
`exa /usr/bin | less`, then quit `less` without scrolling should trigger the following:

```
thread '<main>' panicked at 'failed printing to stdout: Broken pipe (os error 32)', src/libstd/io/stdio.rs:582
```

This is the same issue that `svn log` has (or had; I haven't used svn for a long time).

ogham:
Me neither... fortunately.

However, panicking when printing fails is the correct behaviour (see http://roscidus.com/blog/blog/2013/06/09/choosing-a-python-replacement-for-0install/#safety and #50).

lilydjwg:
You have to find a directory that has many files so that the output can fill up the buffer of the pipe. I have 5062 files under `/usr/bin`.

Printing to a read-only file descriptor will cause a different error:
```
>>> ls 1< /dev/null
ls: write error: Bad file descriptor
```

If the file descriptor refers to a file on which there's no space left, you'll get another error::

```
>>> ls > /dev/full
ls: write error: No space left on device
```

```
>>> ls | echo

>>> echo $pipestatus
141 0
```

Programs exit with an exit code of 141 on SIGPIPE, but no output. You don't expect an error message when you exit your pager, right? When you see it, you have to distinguish where it comes from, the content you're viewing or the program printing?

I don't think a SIGPIPE while printing to a pipe is unexpected; it's expected, indicating the user (or the receiving program) doesn't want (or need) to read more.

ogham:
I made some changes today (turning `println` into `writeln` with a handle that we pass around) and this doesn't seem to come up anymore. Does anyone know if Rust made a change to stop this happening? I'm 99% sure I'm unwrapping all the results.

lilydjwg:
I don't get the error for the command I gave. Then I tried with another:

```
exa -Tl /var | less
```

And I got the error back :-)

gemmarx:
More simply,  on top of exa's git repository

    $ ./target/debug/exa --recurse | head -1

makes a panic as follows
    >>thread '<main>' panicked at 'IO error: Error { repr: Os { code: 32, message: "Broken pipe" } }' .

My execution environment is RHEL 7.2 x86_64 with a dual core CPU.

Cheers.

gemmarx:
If you spoil return value from writeln!,  this trouble goes out of sight.
In lines.rs, grid.rs and details.rs,
```
    - try!(writeln!(....
[...]

PR #118, opened 2016-07-30
Change to ignore broken pipe error

gemmarx:
I think that ogham's opinion in Issue #76 is right.
Panicking when printing fails is the correct behaviour.
However, it is not always a failing that a process as an output destination dies earlier in pipeline.
In most cases, it is just a signal of job termination in traditional unix systems.
Changes of this request make exa ignore broken pipe errors to resolve the issue.


ogham:
Today I get through my exa backlog. Thanks 👍

PR #119, opened 2016-07-30
Change to put error messages into stderr

gemmarx:
Stderr is better than Stdout for printing error messages, isn't it?


ogham:
You’re right, it is!

PR #122, opened 2016-08-29
Add legal values to error messages

quodlibetor:
Now when you do `--sort time` instead of saying "unknown option --sort
time" it will say "unknown options '--sort time' (choices: name...)"
with all legal options.

This also adds the legal values to the default help text.

With this, running exa with bad options looks like:

```
$ target/debug/exa -s blah src
Unrecognized option: '--sort blah'. (choices: name Name size extension Extension modified accessed created inode none)
$ target/debug/exa -lt blah src
Unrecognized option: '--time blah'. (choices: modified accessed created)
$ target/debug/exa --color blah src
Unrecognized option: '--color blah'. (choices: always auto never)
```


ogham:
I like it. Thank you 👍

Issue #135, opened 2016-12-09
Exit with a non-zero status on error.


Stebalien:
Currently `exa /bad/path` exits with a zero exit status. It should really exit with a non-zero exit status.

PR #145, opened 2017-02-26
Update git2 to 0.6.4

spk:
Update git2-rs to 0.6.4 no change needed
Fix https://github.com/ogham/exa/issues/153

PR #146, opened 2017-02-26
Exit with a non-zero status on error

spk:
With `ls` from Debian coreutils 8.26-2

```
ls /bad/path
echo $? # => 2
```

Reproduced same behaviour with exa

Fix https://github.com/ogham/exa/issues/135
Compilation depend on https://github.com/ogham/exa/pull/145


ogham:
Thanks! I can’t believe I missed this out.

spk:
Thanks to you for creating this project and maintaining it !

PR #147, opened 2017-02-26
Use official vagrant jessie64

spk:
https://atlas.hashicorp.com/debian/boxes/jessie64


ogham:
Today I learnt the difference between these two. Thanks again 👍

PR #149, opened 2017-03-30
Update dependencies

Stebalien:
* Also remove bitflags as it isn't used.
* Don't update ansi_term as ANSIString is no longer Clone.

Note: This fixes an openssl build problem I was having (two conflicting openssl versions).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,17 +2,16 @@
 name = "exa"
 version = "0.4.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


emintham:
I had the same issue with building on Ubuntu 16.04. The versions in this MR does fix the issue.

iKevinY:
Interestingly, this problem doesn't seem to occur on macOS.

ogham:
Thanks (y)

ogham:
Thanks 👍

PR #154, opened 2017-03-31
Fix some issues indicated by clippy

neosilky:
Let me know if anything should be removed.


ogham:
Hey, these are great, and absolutely fine. Making the codebase more in line with "common" Rust programming styles is one of the next things I wanted to do, so thank you for your help.

Issue #139, opened 2016-12-25
exa --classify option


grenmester:
There is an option in `ls` called `--classify` which appends an indicator (`*/=>@|`) to each entry depending on the type of entry (directory, binary, symlink, etc.). For example, a `/` would be appended to each directory.

With standard `ls`, you would get something like this:
```
Cargo.lock  Cargo.toml  LICENCE  Makefile  README.md  Vagrantfile  contrib  screenshots.png  src  xtests
```

Whereas with `ls --classify`, you would get something like this:
```
Cargo.lock  Cargo.toml  LICENCE  Makefile  README.md  Vagrantfile  contrib/  screenshots.png  src/  xtests/
```

I think it would be nice if `exa` had this option too.

SShrike:
Just realised this feature was missing, would be nice if it could be added :)

ogham:
Eventually exa's gonna need this for ls compatibility.

ogham:
This got done with #159.

dbeckwith:
This should be added to the README documentation. I was looking to see if `exa` had this and had to search through the issues to find it.

dieggsy:
@dbeckwith you can also always `exa --help`, though I agree it could be added to the README

PR #159, opened 2017-04-13
Implement -F/--classify option

MakeNowJust:
Implement #139. I need this feature.


ogham:
Thanks! 👍

ogham:
Hey @MakeNowJust, in 456fa287d2e2be4f802958c720a7290d162e0855 I added some tests for this, when I noticed that the “block device” and “character device” file types don’t have a `--classify` character. ls doesn’t seem to have characters for them either. Do you think they should? I don’t really use this feature so I’m not sure what would be expected here.

screen shot 2017-04-29 at 09 26 41

PR #167, opened 2017-04-25
Print paths with 2 components properly

### LINKED ISSUE #166: `exa -d foo/bar` just displays "bar"
Running `exa -d foo/bar` just displays the resulting file as "bar". Running `exa -d foo/bar/baz` displays the full "foo/bar/baz".

kballard:
Previously, `exa -d foo/bar` would print the file as "bar", but
`exa -d foo/bar/baz` would correctly print "foo/bar/baz".

Fixes #166.


ogham:
~Who names a variable ‘coconut’?~

Thanks for finding this!

PR #168, opened 2017-04-25
Print . and .. components properly

### LINKED ISSUE #163: `exa -d` doesn't display . for current dir
If I run `exa -d` or `exa -d .`, I get a blank line for output. If I run `exa -ld` or `exa -ld .`, I get a line that shows attributes for a single entry, but with no filename. In both cases, it should show `.` as the filename, like `ls` does.

Example:

```
> exa -d .

> ls -d
./
> exa -ld
drwxr-xr-x - kballard 24 Apr 23:25 
> ls -ld
drwxr-xr-x+ 97 kballard  staff  3298 Apr 24 23:25 ./
```

Reproduced with exa 0.5.0.

kballard:
`Path.file_name()` returns `None` if the path ends in `.` or `..`, which
causes e.g. `exa -d ..` to print a blank line.

Fixes #163.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -102,8 +102,8 @@ impl<'dir> File<'dir> {
 
     /// Create a new File object from the given metadata result, and other data.
     pub fn with_metadata(metadata: fs::Metadata, path: &Path, parent: Option<&'dir Dir>) -> File<'dir> {
-        let filename = match path.file_name() {
-            Some(name) => name.to_string_lossy().to_string(),
+        let filename = match path.components().next_back() {
+            Some(comp) => comp.as_os_str().to_string_lossy().to_string(),
             None       => String::new(),
         };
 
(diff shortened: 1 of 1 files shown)


ogham:
This too! Thanks so much for these.

ogham:
This breaks one of the tests (you know, the ones that didn’t even work properly until yesterday):

screen shot 2017-04-29 at 11 07 27

I’m not even sure if that existing output is correct. The `/` being linked to should be blue because it’s a directory, rather than cyan because it’s a path, right? So there’s definitely *something* up here.

ogham:
second attempt at uploading a screenshot
screen shot 2017-04-29 at 11 07 27

ogham:
I just added a special case when printing these. 956aa85b3b6673a721ff0d3ac122c4111c7d3faa

Issue #134, opened 2016-12-08
Symlink incorrectly shown as broken


nyuszika7h:
I don't know why `exa` thinks this symlink is broken (the second instance of `ghc` in the first example is shown in red and underlined), it works fine.

```
nyuszika7h@cadoth ~ > exa -l /usr/local/bin/ghc
lrwxrwxrwx 24 root 14 Apr  2015 ghc -> ghc
nyuszika7h@cadoth ~ > \ls -l /usr/local/bin/ghc
lrwxrwxrwx 1 root staff 24 Apr 14  2015 /usr/local/bin/ghc -> ../stow/ghc-7.10/bin/ghc
```

ogham:
I'll take a look.

nyuszika7h:
My guess is it tries to resolve `../stow/ghc-7.10/bin/ghc` relative to the current working directory (`/home/nyuszika7h`) instead of the directory of the symlink.

ogham:
@nyuszika7h I think you're right!

ogham:
You were right. It actually took me a while to track down, because it only happens when you specify the file directly on the command-line, rather than specifying a directory. I kept on trying to reproduce it by scanning entire directories, so it never showed up!

But yeah, the problem was with relative symlinks when you were running exa in a different directory.

![screen shot 2017-05-15 at 22 33 06](https://cloud.githubusercontent.com/assets/503760/26080432/81212732-39be-11e7-8101-84b8779d4fa7.png)

PR #169, opened 2017-04-25
Don't prepend current path to symlink targets

kballard:
It's confusing, and `ls` doesn't do this either. We're not prepending the current path to all of the directory entries, and the user is going to interpret the symlink target as relative to the directory containing the symlink.

Also do a little dance to avoid unnecessarily copying the path buffers.

This is going to conflict with #168, but it's a small conflict, and if you merge one of the two PRs I can update the second accordingly.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -169,28 +169,32 @@ impl<'dir> File<'dir> {
             Err(e)    => return FileTarget::Err(e),
         };
 
-        let target_path = match self.dir {
-            Some(dir)  => dir.join(&*path),
-            None       => path
+        let (metadata, ext) = {
+            let target_path_ = match self.dir {
+                Some(dir) if dir.path != Path::new(".") => Some(dir.join(&*path)),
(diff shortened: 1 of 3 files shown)


kballard:
PR updated. I haven't run the xtests yet though.

kballard:
Updated again, xtests now pass.

ogham:
Thanks for this, as always.

In this case, I did have to stare at it for a while to figure out what I thought the correct behaviour should be, and it was this part that did it:

> the user is going to interpret the symlink target as relative to the directory containing the symlink.

All of the other files that we list are relative to the directory that the user provides. I can’t quite remember what my original motivation for doing it the other way was, and I think I was just getting confused between absolute and relative links. Now they’re treated differently, I’m going to add some more tests to fully specify the behaviour, as even now I get it wrong sometimes when I guess what exa’s output is going to be.

It also makes the buggy behaviour in #134 a bit more understandable.

PR #171, opened 2017-04-29
Handle linking to root directories better

kballard:
We don't need a special case for this.

TODO:
- [x] Run xtests


ogham:
Yeah, I see what you’re doing here. The empty `if` did seem a little strange, but I didn’t think of this solution.

PR #174, opened 2017-05-02
Turn off the git networking features

kballard:
This removes openssl and makes the binary properly self-contained.


ogham:
There's no way I'm not merging this

PR #176, opened 2017-05-09
Fix a bunch of issues with the Makefile

kballard:
* `$BASHDIR` and `$FISHDIR` had incorrect paths
* Fix `$FEATURES` to actually work
* Add `-f` to `rm` invocations to quiet missing file errors
* Just use `man` to preview the manpage
* Fix (again) `make` not rebuilding
* Add `make help` to view all of the commands and variables

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 DESTDIR =
 PREFIX  = /usr/local
 
-BASHDIR = /etc/bash_completion.d
+BASHDIR = $(PREFIX)/etc/bash_completion.d
 ZSHDIR  = /usr/share/zsh/vendor-completions
-FISHDIR = /usr/share/fish/completions
+FISHDIR = $(PREFIX)/share/fish/vendor_completions.d
 
 FEATURES ?= default
 
@@ -12,7 +12,7 @@ all: target/release/exa
 build: target/release/exa
 
 target/release/exa:
-	cargo build --release --features "${ENABLE_FEATURES}"
+	cargo build --release --no-default-features --features "$(FEATURES)"
 
 install: install-exa install-man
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,9 +1,9 @@
 DESTDIR =
 PREFIX  = /usr/local
 
-BASHDIR = /etc/bash_completion.d
+BASHDIR = $(PREFIX)/etc/bash_completion.d
 ZSHDIR  = /usr/share/zsh/vendor-completions
```
ogham: Is there a reason `ZSHDIR` doesn’t need `$PREFIX`?

Makefile
```diff
@@ -32,22 +32,43 @@ install-zsh-completions:
...
 clean:
 	cargo clean
 
-
 preview-man:
-	nroff -man contrib/man/exa.1 | less
-
-
-.PHONY: all build install-exa install-man preview-man \
+	man contrib/man/exa.1
```
ogham: I didn’t know you could do this! Just kind of stole [this answer from StackOverflow](http://stackoverflow.com/a/2658839/3484614).

Makefile
```diff
@@ -32,22 +32,43 @@ install-zsh-completions:
...
+
+help:
```
ogham: I really like this help text

Makefile
kballard: I don't use zsh and I'm not sure if it will pick up anything in `/usr/local/share/zsh/vendor-completions`. I did a cursory google search and found other packages explicitly installing into `/usr/share/zsh/vendor-completions`, so it sounds like that's the right place, but someone who actually uses zsh is free to chime in.

kballard:
It occurs to me that this still doesn't use `$DESTDIR` for the completions. I feel like it probably should. Thoughts?

Makefile
ogham: These directories work in the Ubuntu image I use ([the Vagrantfile links them into place](https://github.com/ogham/exa/blob/d2b1499fb12c3fc57041f8eb4d82f615846e5286/Vagrantfile#L72-L79)), but that’s no guarantee that they’ll work in other places, too. None of them work on my computer running OS X.

I tried to find another project’s Makefile that installs shell completions, but came up short.

kballard: By "these directories" do you mean all 3 of them?

Regarding bash, on macOS, `/etc/bash_completion.d` doesn't even exist, and `/usr/local/etc/bash_completion.d` is where various third-party utilities have installed their completions. That said, I don't use bash these days, and poking around in its init files, it actually looks to me as though Bash 3.2.57 on macOS doesn't actually have a default completions directory and users are expected to handle sourcing the files themselves. Either way, third-party software is definitely not supposed to be installing anything into `/etc`, especially when the software itself lives in `/usr/local`.

As for Fish, make sure you're using a recent version. `/usr/local/share/fish/vendor_completions.d` was added in Fish 2.2. Also FWIW, it looks as though Fish's pkg-config actually vends a path for this using `pkg-config --variable completionsdir fish`, so we should consider using that if pkg-config exists, but maybe do that as a followup PR?

ogham: Oh I totally forgot about `pkg-config`! I just tried it on macOS and Linux, which were under `/usr/local` and `/usr/share/local`. Well that’s proof enough that the directories aren’t consistent. That’s something I’d really like to get set up.

exa uses the `bash_completion` package, rather than `bash`, which doesn’t provide enough functionality out-of-the-box. Its FAQ [actually recommends using `pkg-config`](https://github.com/scop/bash-completion#faq) as well, but its directory does default to `/etc`, even though Homebrew (among others, probably) will [move it somewhere else](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bash-completion.rb), probably for the reasons you mention.

I can’t find anything for zsh so it’s looking more and more likely we should just leave it un-prefixed.

kballard: I just filed #179 to track using `pkg-config` for shell completions.

ogham:
Gonna try and get #179 done before 0.6.0 goes out, so the Makefile is complete for it.

kballard:
Just submitted #181 to fix #179.

PR #177, opened 2017-05-10
Fix TextCellContents cjk width

quininer:
before:
screenshot_20170510_162506

after:
screenshot_20170510_162542


ogham:
Good catch. The `DisplayWidth` type was meant to avoid problems like this one, so I have no idea why I didn’t use it! I like the `Sum` implementation, too.

quininer:
Glad you like it. :)

ogham:
I want to merge this in, but your date format change 621bbf73a7a045ae83211946cf540616e34c7a7b has added an extra space to English date formats:

screen shot 2017-05-15 at 22 04 06

Are you using a locale where all the months don’t have the same width, or one with 4-character-wide month names?

quininer:
screenshot_20170516_073051

I think the alignment to 4 is reasonable,
First, it aligns "X月" with "XX月".
Second, it aligns the date with "Date Modified".

ogham:
Ah, I see it now. exa assumes that all month names are going to be the same width. I had a look in my locales directory: `Jan` is the same width as `Dec`, but `1月` is not the same width as `12月`!

However, I would rather not change the width of *all* month names. I think a better solution is to check the width of December at runtime, and use that to determine how wide months should be.

Is this OK? If so, I can merge this PR and make that change.

ogham:
oops wrong button

quininer:
Well, I also prefer an elegant solution.

ogham:
screen shot 2017-05-17 at 22 03 14

The code now checks the width of the 12th month, and uses that value to decide the width.

PR #181, opened 2017-05-16
More Makefile tweaks

kballard:
* Use DESTDIR when installing/uninstalling completions.
* Use pkg-config for FISHDIR/BASHDIR.


ogham:
Functions in Make? What a world. Thanks for taking care of this!

PR #184, opened 2017-05-18
Remove lazy_static crate

kballard:
It's not being used anymore.


ogham:
When I saw the warning against unused `#[macro_use]`, I assumed that I’d just had my Rust compiler updated, and that it was warning against something that used to be fine. Rather than bump the required version of the compiler so close to release (and possibly break something), I just left it in there.

It was only an hour after releasing v0.6.0 that I thought “didn’t I stop using `lazy_static`?”

Don’t code tired, folks.

Issue #185, opened 2017-06-11
locale-0.2.1 does not compile on nightly


gyscos:
`locale-0.2.1` does not compile on rust 1.19 nightly, while `locale-0.2.2` works fine.
The current `Cargo.lock` points to version 0.2.1, and as a result `exa` cannot be compiled on the current nightly.

The fix is simple: run `cargo update` to get the latest locale dependency.

gyscos:
Fixed by #186

PR #186, opened 2017-06-11
Update dependencies

gyscos:
I just ran `cargo update`.


gyscos:
Fixes #185

ogham:
I didn’t notice this broke. Thanks for the fix! 👍

PR #206, opened 2017-08-04
Add snap support for ease of deployment

willdeberry:
I left out instructions for how to install since project owner should be the one setting up Snap Store deployment via https://snapcraft.io/.

I have verified that building the snap locally and installing does work as expected.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,25 @@
+name: exa
+version: 'latest'
+summary: Replacement for 'ls' written in Rust
+description: |
+  It uses colours for information by default, helping you distinguish between
+  many types of files, such as whether you are the owner, or in the owning
+  group. It also has extra features not present in the original ls, such as
+  viewing the Git status for a directory, or recursing into directories with a
+  tree view. exa is written in Rust, so it’s small, fast, and portable.
+
+grade: stable
+confinement: classic
+
+apps:
+  exa:
+    command: exa
+
+parts:
+  exa:
+    plugin: rust
+    source: .
+    stage-packages:
+      - libgit2-24
+      - cmake
+      - libz-dev
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
+parts
+prime
+stage
 target
 
 .vagrant
 ubuntu-xenial-16.04-cloudimg-console.log
+*.snap
--- a/snap/.gitignore
+++ b/snap/.gitignore
@@ -1,0 +1,1 @@
+.snapcraft


snap/snapcraft.yaml
```diff
@@ -0,0 +1,25 @@
...
+summary: Replacement for 'ls' written in Rust
+description: |
+  It uses colours for information by default, helping you distinguish between
+  many types of files, such as whether you are the owner, or in the owning
+  group. It also has extra features not present in the original ls, such as
+  viewing the Git status for a directory, or recursing into directories with a
+  tree view. exa is written in Rust, so it’s small, fast, and portable.
+
+grade: stable
+confinement: strict
```
OddBloke: Won't the `strict` confinement here mean that I can't use `exa` against anything but the files that are shipped with the snap. For use as a general system utility, I would have expected this to be a classic snap.

OddBloke:
(Once this is merged, I'd strongly recommend using https://build.snapcraft.io to get automatic builds from master.)

snap/snapcraft.yaml
willdeberry: How are you thinking you want to use `exa` against things outside the snap? If you are concerned with running `exa` in general, all seems well on my end. I am happy to test functionality for you with the snap installed to see what the limitation are.

OddBloke: I'm failing to build locally, so I can't test for myself, unfortunately. How are you installing the snap? strict snaps shouldn't have access to any files outside of the snap confinement (but if you've installed it in devmode, you wouldn't see this behaviour).

To confirm this, I created an `ls` snap and see this behaviour:

```
$ /snap/bin/ls
/snap/ls/x1//bin/ls: cannot open directory '.': Permission denied
$ /snap/bin/ls /
/snap/ls/x1//bin/ls: cannot open directory '.': Permission denied
```
but accessing its own files will work:

```
$ /snap/bin/ls ~/snap/ls
common	current  x1
$  /snap/bin/ls /snap/ls
current  x1
```

If you change confinement to "classic", then exa will be installed in a way that allows it full access to the system.

willdeberry: You were correct and I was using devmode. I am still learning but saw this project on reddit and figured I'd assist. I've adjusted the confinement in the latest commit/push.

OddBloke:
Looks good to me!

ogham:
I’ve been looking into Snap; I’d like to have a better building/distribution mechanism for the release (in a couple days)

willdeberry:
Just need to hook into https://build.snapcraft.io and it will build a new snap for any push to master branch just like @OddBloke stated. I utilize this for one of my projects and it only takes a couple minutes to setup.

tetienne:
Any update on this PR? Will be so nice to have exa available with snap.

HermannBjorgvin:
Any update on this? This is a good feature for installation on debian/ubuntu.

ogham:
I have not gone through the steps of setting up Snap yet, but this file is now at least merged! Thanks 👍  and sorry it's taking so long

tnoskilz:
any update on getting it published to snap store?

k8s-gpubuilder:
2 years later, is this finally dead in the water?

ariasuni:
The attention I wish this project had would be more contributors but eh. exa isn’t dead though, I still put a lot of efforts in triaging issues. That’s actually most of my time in the project, and why open issues don’t accumulate ridiculously and stay near the 200 count.

But I guess this was mostly addressed to @ogham; not sure when he’s gonna work on this project again, and I don’t feel comfortable forking the thing because I wouldn’t be able to manage everything on my own.

k8s-gpubuilder:
Well, if you want to use exa (i.e. install) that is currently a pretty big pain, since the most used distros (ubuntu 20.04), debian and the most used package systems are not even supported. I think that might be a reason for people not using it. I can't currenly use it, as we have a company policy to only use package managed software that is cleanly uninstallable, so I can't just copy over the executable. 

I applaud your efforts, I'm excited to replace ls, as it is lacking in many many ways and I'm an innovator that is usually far ahead of the curve and picks up many years before mainstream does. (I used replaced top with htop because it was just better, a thing that took mainstream users 5 to 10 years longer). 
I guess ls is so known that many people don't even question it, that makes exa a harder sell. It is a shame that the installation process is difficult and probably too much for most users. A snap would be great in my view, as it provides a way to install it on any Linux distro, no matter what version, but unfortunately I can't provide help packaging it, as I am super busy right now.

PR #219, opened 2017-08-04
Fix typo in filename

xyproto:
* And update the year in the `LICENSE` file to 2017


ogham:
Sorry, that's just how I spell it! I'm not aware of any automated scripts or things that rely on it being named like that, other than the GitHub one, and that can detect it already.

(Should probably update the year though.)

PR #224, opened 2017-08-04
Update filetype.rs

pskuza:
Added a few video and audio extensions


ogham:
Neat. I'm not 100% about `.ts` because I've seen it used for TypeScript, but I'll include it anyways.

hoodie:
`.ts` is most definitely typescript

pskuza:
.ts can also be transport stream which is common for linux setup boxes that record tv stations, you'll end up with a .ts file.

hoodie:
Right, I kinda assume more of exas users having to do with typescript then with that format.

PR #225, opened 2017-08-04
Add quick-and-dirty Debian package spec for cargo-deb

ittner:
That's not a package you can show to your mother, but it makes easier
to install/update/uninstall exa in systems without cargo or rustc.


ogham:
(2½ years later) Thanks!

PR #237, opened 2017-08-05
Use term_size crate to fetch terminal size.

PJB3005:
Should be cross-platform, one step closer to Windows support.


ogham:
Oh cool, I didn't know this existed. Thanks :thumbsup:

ogham:
^ that was a bit of a head scratcher tho

PR #239, opened 2017-08-05
Better date and time format

whuji:
For french users, the month with the longest short name is not December,
it is July (juil. vs dec), and the maximum length of a short name month
is not 3, nor 4 but 5.


ogham:
Thanks for submitting this PR. Tests with different locales are something I still need to add.

Unfortunately I can't merge this in because it would re-introduce the issue from #177, which was the one that introduced the December check in the first place. The best thing to do to avoid this type of bug in the future would be to loop through the month names at runtime and determine the maximum width of them all. I raised #244 to track that.

Issue #244, opened 2017-08-05
Fix month name widths once and for all


ogham:
#177 fixes the month names for December, which was the longest in Japanese. #239 fixes them for July, which was the longest in French.

To avoid this sort of bug in general, exa should find out the month with the longest name among *all* months, and use the width of that.

whuji:
I did not see the issue in japonese. I have modified the patch to implement the loop over the month names.

ogham:
Closed by #247

PR #247, opened 2017-08-06
Fix month name widths once and for all #244

whuji:
To render the date, Exa now find out the month with the longest name
among all months, and use the width of that.

--

I think that adding lines in the match to call DateFormat::parse with the good format is ugly, but as the datetime library does not accept a String (only a &str), and as I do not know Rust, I did not find a more elegant way. Well, actually, it works !

### DIFF
--- a/src/output/time.rs
+++ b/src/output/time.rs
@@ -1,6 +1,7 @@
 use datetime::{LocalDateTime, TimeZone, DatePiece, TimePiece};
 use datetime::fmt::DateFormat;
 use locale;
+use std::cmp;
 
 use fs::fields::Time;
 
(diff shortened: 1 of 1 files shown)


ogham:
Thanks for making that change! You basically wrote it the same way I would have, don't worry. I know about the restriction in rust-datetime, and yeah, it is annoying.

GreenRecycleBin:
I'm assuming this restriction is due to [`DateFormat`s having `static` lifetime in exa](https://github.com/ogham/exa/blob/dd0c5f6bef10864f94b4e3388b212499a52fc963/src/output/time.rs#L47-L51) and the current [`DateTime::parse` implementation](https://github.com/rust-datetime/datetime/blob/e5857d960310baa3fa7d6b01a67c7ec8d8142e1c/src/cal/fmt/custom.rs#L168-L173). In `pub fn parse(input: &'a str) -> Result<DateFormat<'a>, FormatError>`, `input` has the same lifetime as that of `DateFormat` result. Hence taking a reference of a local `String` (e.g. a format string constructed programmatically) won't work, since the `String` doesn't have `static` lifetime.

@ogham Could you share more about the reasons for this limitation? I'm new to Rust and I'm eager to learn more. Thank you.

PR #249, opened 2017-08-06
Fix typo

zeeshanu:


ogham:
Not a typo! `rustc` is the Rust compiler executable.

Issue #213, opened 2017-08-04
'iso' date, using MM-DD HH:MM for previous year


uazu:
Using `--time-style=iso`, the current year is displayed as YYYY-MM-DD, and previous year as MM-DD HH:MM. Surely it should be the other way around?

ogham:
Yes, it should! I think this is pretty clearly a programming mistake.

ogham:
Fixed by #250 and tested by 4cb28593fdd13a0c6269cb6e98332bd9e9ce815a.

PR #250, opened 2017-08-06
fix issue #213 iso date format inversion

whuji:


ogham:
Thanks for taking care of this one too.

Issue #233, opened 2017-08-04
Different color shades for Date column, like `lsd`


huxi:
Like `--color-scale`, but for the Date column.

#### Date levels
- today
- yesterday
- this week
- this month
- this year
- previous years

ogham:
Good idea

excalibur1234:
i also like this idea.

in my opinion, k already does this really well, just in case some inspiration is needed:
https://github.com/supercrabtree/k#rotting-dates

lilyball:
https://github.com/Peltoche/lsd does this by default; it appears to color files from last year separately as colors from this year, and recent files (maybe past hour?) separately as well. I also like the colors it's chosen for this.

ariasuni:
This feature has been added in [the active fork eza](https://github.com/eza-community/eza).

PR #253, opened 2017-08-06
added colorizing for date time recent

emattiza:
feature addition to render pipeline for time or date related fields to be gradient scale (dark blue: recent to light blue: old). See ogham/exa#233

see below image. some folder names redacted for my privacy, but this is example output

exa-color-time


excalibur1234:
this is really cool.

is it possible to also use grey (from white to almost black) instead of blue (cyan to purple) like in k? https://github.com/supercrabtree/k#rotting-dates

ogham:
I like this idea! The colours can coexist, too: file sizes go green to orange and timestamps go blue to purple.

However, like the scale for file sizes, I'd rather this exist behind an option rather than be enabled for everybody by default. Is that alright? If so, I could merge it in then take care of the options and tests and stuff.

emattiza:
The only difficulty I've had is getting gradients out of 256 xterm colors. I'd appreciate help on what exact colors are the best choice to pass to fixed for this behavior.

+1 on optional flag for verbose coloring.

ogham:
When I picked the ones for file sizes, I just looked at a picture that had all 256 xterm colours in it, found some that looked right, then tested them with some terminal colour schemes to make sure they worked well enough with light/dark/neutral backgrounds. That was it!

Anyway, I'll try them out later when I work on this, and if any don't match I'll let you know.

timofonic:
@excalibur1234  Thanks!

@ogham Will this be merged someday? Please...

emattiza:
@timofonic I'll give a rebase of the PR to master and look to implement the option tonight if i can find time.

emattiza:
yeah i didn't see you implemented LS_COLORS and EXA_COLORS support... 

I'll look into finding something compliant. Would it be better if EXA_COLORS had custom colors for each of the time scales suggested?

emattiza:
Leaving this for review! It's working, but I feel it's a bit hacky. I would love some assistance "rustifying" the code a bit. I am pretty unfamiliar with the language and may be missing some helpful patterns.

doy:
hey, curious what the status of this is - i'd also like this functionality!

src/style/colours.rs
```diff
@@ -386,6 +416,30 @@ impl render::SizeColours for Colours {
...
+              self.date.time_month
+          } else if age < 60 * 60 * 24 * 365 {
+              self.date.time_year
+          } else {
+              self.date.time_past
+          }
+      }
+      else {
+          self.date.time_today
+      }
```
ariasuni: I think it would be better to use named colors, as I did here:
https://github.com/ogham/exa/pull/370

src/style/colours.rs
```diff
@@ -179,9 +193,18 @@ impl Colours {
...
 
+            date: Date {
+                time:           Blue.normal(),
+                time_today:     Fixed(93).normal(),
+                time_yesterday: Fixed(99).normal(),
+                time_week:      Fixed(105).normal(),
+                time_month:     Fixed(111).normal(),
+                time_year:      Fixed(117).normal(),
+                time_past:      Fixed(123).normal(),
+            },
```
ariasuni: I think it would be better to use named colors, as I did in my PR.

ariasuni: I think it would be better to use named colors, see #370.

jasonkena:
status?

ogham:
Again, apologies for dragging my feet on this PR. I ended up merging a different date-time-recent branch, so I'm going to document that one instead of merging this.

PR #258, opened 2017-08-07
Add extra types to compressed filetype filter.

chrisvittal:
This adds a few more common compressed filetypes to the is_compressed
fuction. Notably, xz, and two common package file formats, deb and rpm.

Arch Linux's default `dircolors` sets all of the following extensions to be colored with `01;31` for bold red. They represent compressed files or other archives. They are `tar tgz arc arj taz lha lz4 lzh lzma tlz txz tzo t7z zip z Z dz gz lrz lz lzo xz zst tzst bz2 bz tbz tbz2 tz deb rpm jar war ear sar rar alz ace zoo cpio 7z rz cab wim swm dwm esd`. I thought that I would bring exa more in line with this as it was surprising when an `xz` archive was not colored.


ogham:
Thanks! 👍

Issue #284, opened 2017-09-10
Default -t to modified


mqudsi:
[For compatibility with ls](http://man7.org/linux/man-pages/man1/ls.1.html), change `-t` from "optional flag with required value" to "optional flag with optional value" defaulting to modification date.

mqudsi:
So this is apparently less straightforward than I had realized. I thought `--modified` sorts by modification date, but that only sets the behavior of the timestamp column.

How insane would it be for `exa -t` to be interpreted as `exa -s modified` in the name of ls compatibility? This doesn't clash with anything or break backwards compatibility with previous exa syntax, but it would be odd.

ogham:
I did some work on this last night before I saw your second message.

Unfortunately, there's no way to have an optional argument in exa: either it always takes a value or it always doesn't. This is because the options parser allows you to drop the equals for single-letter options. If you ran:

    $ exa -ltr

Then it'll parse it like this instead, because `-t` takes a value:

    $ exa -l -t=r

So what exa now does is try to detect this case (by looking at the error messages) and guess which command you intended to run, based on what `ls` would do:


It only works for those two examples, but it's *something*.

mqudsi:
Perhaps it would be best to add a --legacy flag to exa with ls-compatible flags. Then users could alias `ls` to `exa --legacy` and that would make everyone happy.

ogham:
If I did that, I'd check the name of the running program to see if it's `ls`. This is similar to how old editions of `true` and `false` used to work: they were the same binary that checked itself to see what it should return (IIRC, anyway)

ogham:
Anyway, exa and ls are very similar, so subsuming all of ls's functionality isn't too far-fetched an idea!

andrew-lee-metaswitch:
The [release notes](https://github.com/ogham/exa/releases/tag/v0.10.0) suggest this is fixed, but that's not my reading of the [associated commit message](https://github.com/ogham/exa/commit/36cf5df044cea9f89bb9569cb6a42f072c12fdea ).

ariasuni:
Yep, it’s an error that this issue was closed. But it’s tracked at #519 anyway, so let’s keep this one closed.

PR #289, opened 2017-09-16
Optional args support

mqudsi:
I added support for optional arguments that take optional values to exa's command parser when I was working on #284, although that particular fix didn't materialize for reasons explained in that thread.

I figured I'd open a PR just in case you wanted this so you don't have to rewrite this in the future. Feel free to reject.

Adds support for `--flag[=optional_value]` to the parser, so that a flag modifying the behavior of exa can be specified, defaulting to a certain behavior which can be modified by supplying `optional_value`.

As explained in 36cf5df044cea9f89bb9569cb6a42f072c12fdea, this is equivalent of getopts' `optflagopt` mode.

### DIFF
--- a/src/options/parser.rs
+++ b/src/options/parser.rs
@@ -92,6 +92,9 @@ pub enum TakesValue {
 
     /// This flag will throw an error if there’s a value after it.
     Forbidden,
+
+    /// This flag may be followed by a value to override its defaults
+    Optional,
 }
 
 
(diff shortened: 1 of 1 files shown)


timofonic:
@mqudsi That's very nice! Thanks a lot!

@ogham Any hopes to merge this? The project seems stalled :(

ogham:
So I did finally merge this, after all this time.

Here's what happened: I didn't want exa to have this feature, because I wanted the command-line options to be completely consistent, but I couldn't find a way to write out the reason why in this comment box. Now I'm coming back to the code, I've changed my mind about that — exa already has plenty of shortcuts, and I've seen the same behaviour in other tools — so I guess the reason was inside me all along.

mqudsi:
Hey, you never closed it all this time. That's something.

PR #290, opened 2017-09-16
Add make test to Makefile

mqudsi:
Runs `cargo test ...` when `make test` is called.


jbergstroem:
For old farts like me, could you perhaps alias `check` to `test`? 😄

mqudsi:
Added.

ogham:
Merged! Thanks.

Sorry this one sat on the side for so long.

PR #300, opened 2017-10-04
Fix LS_COLORS 01 and 04 detection

chetgurevitch:
Necessary to correctly parse LS_COLORS as generated by dircolors.

### DIFF
--- a/src/style/lsc.rs
+++ b/src/style/lsc.rs
@@ -52,8 +52,8 @@ impl<'var> Pair<'var> {
             match num {
 
                 // Bold and italic
-                "1"  => style = style.bold(),
-                "4"  => style = style.underline(),
+                "1" | "01" => style = style.bold(),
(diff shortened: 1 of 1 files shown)


hoodie:
this could be put on the same line to avoid confusion:

```rust
                 "1" | "01"  => style = style.bold(),
                 "4" | "04"  => style = style.underline(),
```

chetgurevitch:
Done

clarcharr:
Is there a reason why the code is just directly matching strings instead of just parsing ints?

ogham:
Nice, thanks. 300 🎉

I'm not actually sure whether the ANSI codes are meant to be strings of digits or just integers. I just went with strings because I didn't think about trying to line the numbers up with one another.

ogham:
I released this as a weekly if anyone wants this behaviour early: https://bsago.me/blog/exa/v0.9.0-weekly-1

PR #302, opened 2017-10-05
Update README with new cargo instructions

gyoshev:
After 1efb8a5113f1a5ffeb0af6fe4e7374502ac41712 the binary can be installed without specifying the git repo.


ogham:
Thanks! 👍

PR #304, opened 2017-10-10
Add --classify to man page

polyzen:


ogham:
I could have sworn that was in there already! Thanks.

PR #343, opened 2018-01-30
Add new sort option `.name` and `.Name`

BenWhitehead:
I've really enjoyed using exa for a few months now, but there was one sorting
option that I did miss from my previous ls config. The ability to have sort ignore
if a file was "hidden" or not when sorting.

Add two new sort options `.name` and `.Name` which with ignore a leading
`.` if present on the file name before sorting according to `name` and
`Name`.

This new sort is convenient if you want to list hidden and unhidden
files sorted together.

I added a couple tests to check the parsing of the new args works correctly,
but I didn't figure out how to add a new test where a list of files rendered and
compared to an expected output (apologies if this should be obvious, I'm still
pretty new to rust and learning the ropes).

Any feedback or pointers are welcome.

### DIFF
--- a/src/fs/filter.rs
+++ b/src/fs/filter.rs
@@ -191,6 +191,10 @@ pub enum SortField {
     /// bad, even though that’s kind of nonsensical. So it’s its own variant
     /// that can be reversed like usual.
     ModifiedAge,
+
+    /// The file's name, however if the name of the file begins with `.`
+    /// ignore the leading `.` and then sort as Name
+    NameMixHidden(SortCase),
 }
 
 /// Whether a field should be sorted case-sensitively or case-insensitively.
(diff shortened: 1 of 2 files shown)


ogham:
This is a really neat idea! It’s a useful feature, it slots right in to the `--sort` argument, and I’m sure at some point in the past I’ve actually needed to do this. I also don’t see any problems with your Rust code in this PR. Definitely merged. 👍 

Can I ask how you were able to do this with ls? I’m aiming to (eventually) cover all the features of common ls implementations.

> I added a couple tests to check the parsing of the new args works correctly,
but I didn't figure out how to add a new test where a list of files rendered and
compared to an expected output (apologies if this should be obvious, I'm still
pretty new to rust and learning the ropes).

exa does have integration tests like these, but instead of using Rust like the unit tests, it has a Vagrant image and some shell scripts to run the `exa` binary repeatedly. I’ve written some documentation for this but it’s not public yet, so don’t worry about doing this yourself.

BenWhitehead:
I'm not actually sure how this happened in `ls` to be quite honest, I remember seeing it on an openSuSE Linux server several years ago and have missed it on my Debian based machines since then.

Thanks for accepting my PR!

PR #344, opened 2018-01-31
remove root section from cargo.lock

rleungx:
rust-lang/cargo#4563
rust-lang/cargo#4571


ogham:
Thanks! I hadn't kept up with the Cargo changes and didn't get why the lockfile format changed. (Earlier PRs made this a pain to merge)

PR #345, opened 2018-01-31
remove unused import

rleungx:
It seems that AsciiExt is not used.


hoodie:
you are probably using rustc >= 1.23 which [deprecated this module](https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries). For compatibility with previous versions of the compiler this should probably stay there. Though I'm not quite sure how the lib team suggests to handle this.

rleungx:
@hoodie You are right and maybe we can use #[allow(unused)] here.
rust-lang/rust#46510 mentioned this.

ogham:
Thanks for this! I did wonder why a warning suddenly appeared one day. I'm going to let rustc 1.23 be a little more widespread before removing the import entirely.

PR #355, opened 2018-02-14
Create worker threads only once (for performance)

alexmaco:
This improves performance by a factor of at least 2 in large --tree workloads by avoiding the repeated creation/destruction of the pool and containing threads.
Cycling pools also encountered lots lock contention, which accounted for most of the time saved by reusing a single pool.


ogham:
This is great! I ran a benchmark and it was significantly faster (87s vs 109s to scan all of `/`). Thanks for the PR 👍

Issue #346, opened 2018-02-02
Can you also highlight lowercased `readme.md` files?


nikitavoloboev:
`exa` highlights in yellow files like `README.md`. But it doesn't highlight `readme.md` file (when it is lowercased).

Is it possible to add support for it too?

joshleeb:
I think this would be nice. I often see projects with lowercase `readme.*` files.
I'll put up a quick PR for this.

ogham:
(Looks like this didn’t auto-close)

PR #356, opened 2018-02-15
Case Insensitive README.* Files As immediate

joshleeb:
Ref. #346.

Update the `immediate` classification of a file to include files with the case insensitive prefix `readme`. This means that lowercase `readme.*` file will be highlighted.


ogham:
I don't think I've ever seen a lowercase readme, but I'm sure they exist! Thanks.

dguo:
They do! Sindre Sorhus (who has made hundreds of Node packages), [lowercases all of his](https://github.com/sindresorhus/ama/issues/197).

PR #366, opened 2018-03-15
Don't treat .ts as a video file extension

dguo:
It was originally added in #224, where the TypeScript concern was already brought up. Given the target audience of `exa`, my guess is that it's more likely the user will be dealing with TypeScript files rather than `.ts` video files. In the most recent [Stack Overflow developer survey](https://insights.stackoverflow.com/survey/2018#most-loved-dreaded-and-wanted), TypeScript was the number 4 most loved language and number 5 most wanted. It's hard to draw conclusions from that, but I think it's fair to say that it's not an obscure technology.

Thanks for making `exa`, by the way! I love the colors. :rainbow:

### DIFF
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -39,7 +39,6 @@ impl FileExtensions {
         file.extension_is_one_of( &[
             "avi", "flv", "m2v", "mkv", "mov", "mp4", "mpeg",
             "mpg", "ogm", "ogv", "vob", "wmv", "webm", "m2ts",
-            "ts",
         ])
     }
 


ogham:
I was hesitant to change an older PR like this, but I’ve started using TypeScript for work and I’ve been seeing this happen too. So I’m going to merge this. It’s annoying because I also see `.ts` for video files a lot, and I wasn't really sure which one is used *more*; I do agree that the developer survey lends some weight to TypeScript being popular though.

In the end, it’s better to incorrectly *not* highlight a file than to incorrectly highlight one. It’s possible to get the old behaviour back by setting an environment variable:

    EXA_COLORS="*.ts=38;5;135”

Glad you like exa!

Issue #375, opened 2018-04-11
Git status does not work for parent directories


LlinksRechts:
If one is in a subdirectory of a git repository, `exa -l --git ../` does not display the right git status for the files in the parent directory

```
[/some/repo] $ exa -l --git
drwxr-xr-x    - user 11 Mar 23:25 -M directory
.rw-r--r--    0 user 11 Apr 15:58 -N file
[/some/repo/directory] $ exa -l --git ../
drwxr-xr-x    - user 11 Mar 23:25 -- directory
.rw-r--r--    0 user 11 Apr 15:58 -- file
```

PR #378, opened 2018-04-13
fixed git status not working in directories containing '../'

LlinksRechts:
As suggested in issue #375 a path containing ../ would resolve to one which can not be parsed by the git library (eg. `../child` resolves to `/some/directory/parent/child/../child` instead of `/some/directory/parent/child`), leading to the git status always displaying `not modified`; this is fixed in this pull request


ogham:
Thanks!

PR #398, opened 2018-06-19
fix most clippy warnings

thibran:


ogham:
Ah, clippy. You might be able to tell how old some of this code was!

I'm going to merge this (and deal with the conflicts), but there's one place where I think clippy has gone too far, and that's replacing `new` with `Default::default` for `ISOFormat` and `DefaultFormat`. Constructing these two involves looking up either the current date, the user's current locale, or both, and I think there's an implicit assumption that using `default` will produce the same value everywhere. A better solution imho would be to rename it to `load` instead, because then the name tells you it's loading something, rather than just constructing a value.

Anyway, thanks for going through the warnings. The ones about `std::ascii` in particular were annoying me the other day.

thibran:
That's nice to hear – very kind from you to resolve the merge conflicts.

PR #405, opened 2018-07-13
Flag for only showing directories

### LINKED ISSUE #401: Flag for only showing directories in results
Hi! Seems like `exa` has a bunch of flags similar to `tree`, but one feature I found myself wanting recently is `tree`s `-d` flag for only showing directories. It looks like `exa` has `-d` for `--list-dirs`. Would it be useful to others to have a similar, but obviously differently named, flag to only show directories in the results? Perhaps there is already a way to do this already with `exa?

Thanks!

Don42:
Implement a filter and associated short and long flags ('-D', '--only-dirs')
to only show directories. The goal was to get this to work in the tree
view, but it was easiest to implement it globally.

Closes #401

### DIFF
--- a/src/fs/filter.rs
+++ b/src/fs/filter.rs
@@ -41,7 +41,7 @@ pub struct FileFilter {
     /// ones, depending on the sort field.
     pub reverse: bool,
 
-    /// Whether to only show directories
+    /// Whether to only show directories.
     pub only_dirs: bool,
 
     /// Which invisible “dot” files to include when listing a directory.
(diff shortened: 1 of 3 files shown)


src/fs/filter.rs
```diff
@@ -41,6 +41,9 @@ pub struct FileFilter {
     /// ones, depending on the sort field.
     pub reverse: bool,
 
+    /// Whether to only show directories
```
mc10: Nit: this should have a period for consistency.

Don42: Should be fixed now.

ogham:
> The goal was to get this to work in the tree view

You know, I’ve never really seen the use for wanting to see only files or only directories, but I didn’t think how it would look in the tree view. Thanks!

PR #413, opened 2018-09-09
update libc to latest version

semarie:
needed for include arm64 definitions for OpenBSD

without it, `exa` doesn't compile at all (failed to build `libc` dependency) on OpenBSD on arm64 architecture.


ogham:
Thanks!

Issue #417, opened 2018-10-02
No support for .m4v files


Fierthraix:
If you have an example folder of videos of various formats (as shown here):
![exa1](https://user-images.githubusercontent.com/27262868/46331384-a74faf00-c5e5-11e8-8893-be7e2bc81640.png)
The `.m4v` file doesn't have the purple colour of all the other video types.

Although the argument could be made that M4V files shouldn't be recognized as videos because they don't have a [media type given out by IANA](https://www.iana.org/assignments/media-types/media-types.xhtml), M4V is a widely recognized type. It has both a [Wikipedia](https://en.wikipedia.org/wiki/M4V) and a [fileinfo](https://fileinfo.com/extension/m4v) page as well as wide support in [many media players and Apple](https://www.techopedia.com/definition/10715/m4v).

Fierthraix:
Fixed with https://github.com/ogham/exa/pull/418

PR #418, opened 2018-10-02
Add support for Apple's M4V format

Fierthraix:
Closes https://github.com/ogham/exa/issues/417

This commit makes it so `.m4v` files are recognized as proper video files.
exa2

The first command is before the commit, and the second is from after.
All tests pass.


ogham:
Thanks!

PR #421, opened 2018-10-05
Documented nix availability

mgttlinger:
`exa` is available in the `nix` package manager. Added that info to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,11 @@ or:
 
 [Formulae](https://github.com/Homebrew/homebrew-core/blob/master/Formula/exa.rb)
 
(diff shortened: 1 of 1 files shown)


ogham:
Good to know!

GladOSkar:
Also available in Solus (`eopkg it exa`) - Should all distros that offer `exa` be added?

ogham:
@GladOSkar I hadn’t heard of Solus, but I don’t see why not. Is that the installation command? I can add it to the README.

GladOSkar:
@ogham yes it is

PR #430, opened 2018-10-21
Added Rationale section to README.md (fixes #403)

### LINKED ISSUE #403: Document rationale in README?
Here are some example questions that would be cool to see answered in the README file:

- What were the motivations for a `ls` replacement?
- Where does `exa` shine where `ls` doesn't?
- What are `exa`'s goals? (to be the fastest? the easiest to use? etc)

pconcepcion:
Trying to explain the motivation and advantages of exa as requested on the issue #403

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## Rationale
 
-**exa**  is a modern replacement for the command-line program ls that ships with Unix and Linux operating systems, is a modern file lister with more features and better defaults. It uses colours to distinguish file types and metadata. It knows about symlinks, extended attributes, and Git. And it’s **small**, **fast**, and just one **single binary**.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,12 @@
 
 [exa](https://the.exa.website/) is a replacement for `ls` written in Rust.
 
+## Rationale
+
+**exa**  is a modern replacement for the command-line program ls that ships with Unix and Linux operating systems, is a modern file lister with more features and better defaults. It uses colours to distinguish file types and metadata. It knows about symlinks, extended attributes, and Git. And it’s **small**, **fast**, and just one **single binary**.
```
ariasuni:
```suggestion
**exa**  is a modern replacement for the command-line program ls that ships with Unix and Linux operating systems, with more features and better defaults. It uses colours to distinguish file types and metadata. It knows about symlinks, extended attributes, and Git. And it’s **small**, **fast**, and just one **single binary**.
```

ogham:
This works nicely!

PR #432, opened 2018-10-26
Documentation fixes

jcrd:
Moves --level option to the filtering group in help and manpage and documents --only-dirs (#405).


ogham:
Thanks!

PR #440, opened 2018-11-14
Exclude unneded files from crates.io

ignatenkobrain:


ogham:
Thanks, I didn’t know you could do that.

PR #443, opened 2018-11-20
Use more accurate time (fix inconsistent sorting)

### LINKED ISSUE #411: Sorting by time is inconsistent
Hi, I've found that the time sorting feature isn't consistent.
It seems to sort differently on each execution of the command -  for files which were last modified at the same time.

I've noticed it while using the command 'watch'.

The command I ran:
watch -n0.1 'exa -mbghHla --SORT=time |tail'

I tried to.execute the same command but this time I used 'ls' and it worked flawlessly.

(The ls command - watch -n0.1 'ls -ltr | tail')

ariasuni:
Fix #411 

Sort based on `mtime` and `mtime_nsec()` instead of `mtime()` only, same for `atime()` and `ctime`.

### DIFF
--- a/src/fs/fields.rs
+++ b/src/fs/fields.rs
@@ -167,7 +167,7 @@ pub struct DeviceIDs {
 
 
 /// One of a file’s timestamps (created, accessed, or modified).
-#[derive(Copy, Clone)]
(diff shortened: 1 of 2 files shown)


ogham:
Brilliant, thank you.

ogham:
(I swear `modified_time` etc weren’t there when I last looked)

PR #446, opened 2018-11-28
Format with cargo fmt

joshleeb:
Apologies for the nit PR, but this PR just runs `cargo fmt`.

A majority of other rust projects I've seen with any amount of contributors has all rust code formatted before it's committed. I thought it would be worthwhile to do the same for `exa` (and also so I can keep `format-on-save` on when I'm working on `exa` 😝).

Once this is merged, I'd be happy to put up another PR that checks formatting in the travis builds, perhaps as an "allowed to fail" build, or perhaps not (but I thought that would be good to have in a separate PR)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,10 +9,9 @@
 ///
 /// - https://stackoverflow.com/q/43753491/3484614
 /// - https://crates.io/crates/vergen
-
 extern crate datetime;
-use std::io::Result as IOResult;
 use std::env;
+use std::io::Result as IOResult;
 
 fn git_hash() -> String {
     use std::process::Command;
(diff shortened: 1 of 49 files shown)


ogham:
Hey, thanks for the PR (nit PRs are totally valid!), but I’m not interested in auto-formatting the code yet.

Merging it would have an effect on all the other pull requests (of which there are quite a few, I’ve let the list build up a little). But also, I don’t fully agree with some of the decisions that rustfmt makes. There are places in the code where I’ve purposefully inserted whitespace to make different parts of the code line up, and formatting takes the spacing right out. I know that one day I'll have to grit my teeth and start putting the `else` on the same line as the `}` and that sort of thing, but until I’ve found the right combination of rustfmt flags, I’m keeping the code hand-formatted.

joshleeb:
No problem - that seems completely fair 😄

Thanks for taking the time to explain that!!

PR #450, opened 2018-12-07
Update to Rust 2018

Bond-009:
Brief summary:
* Extern crates are no longer needed.
* Absolute paths begin with a crate name, where the keyword crate refers to the current crate.
* Import macros with use, rather than a `#[macro_use]` attribute.

### DIFF
--- a/src/options/view.rs
+++ b/src/options/view.rs
@@ -1,20 +1,21 @@
-use output::{View, Mode, grid, details, lines};
-use output::grid_details::{self, RowThreshold};
-use output::table::{TimeTypes, Environment, SizeFormat, Columns, Options as TableOptions};
-use output::time::TimeFormat;
+use lazy_static::lazy_static;
 
-use options::{flags, Misfire, Vars};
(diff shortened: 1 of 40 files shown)


src/options/view.rs
```diff
@@ -1,19 +1,22 @@
-use output::{View, Mode, grid, details};
-use output::grid_details::{self, RowThreshold};
-use output::table::{TimeTypes, Environment, SizeFormat, Columns, Options as TableOptions};
-use output::time::TimeFormat;
+use lazy_static::{lazy_static, __lazy_static_create, __lazy_static_internal};
```
ariasuni: I feel you should use `use lazy_static::*;`, because `__lazy_static_create` and `__lazy_static_internal` are implementations details of lazy_static.

ariasuni:
Since a few PR were merged just now, you should to rebase your changes to resolve conflicts — if you’re still wanting to work on that.

ogham:
Thank you! 👍

Issue #434, opened 2018-10-29
Flag/colour for directory which is a btrfs subvolume


daviessm:
Would it be possible to flag a directory which is a subvolume in the `btrfs` filesystem? These directories have an inode type of 2 or 256 when the filesystem type is btrfs: https://stackoverflow.com/a/32865333/5293556

I wouldn't mind whether a character other than `d` or a different colour was used for these, it would just be nice to see them.

daviessm:
Upon further investigation, it's the inode *number* which signifies a subvolume, so I can see them by using `-i` and looking for `2` or `256`. It would still be nice to highlight those without having to look through inode numbers though.

daviessm:
I have started work on this in https://github.com/daviessm/exa/tree/issue_434_subvolumes - it correctly detects subvolumes by ino as described above, but needs to also check that the filesystem type is btrfs. 

This can be done with `getmntent()` from libc as described here: https://stackoverflow.com/questions/40659638/how-can-i-determine-the-mount-path-of-a-given-file-on-linux-in-c but that will need to be cached as there could be thousands of calls per run of `exa`.

ogham:
Hi there, and sorry for not leaving a reply until now.

I've looked at your code, and it all looks good. If detecting a subvolume really is as easy as checking the inode, then I'll be happy to merge it in.

daviessm:
No worries, yes it is that simple - but obviously this will only apply if the volume the directory is on is actually btrfs - otherwise there's a remote possibility of false positives (where the inode number of a directory just happens to be 2 or 256). I haven't figured out yet how to check that.

ogham:
Aaah, you’re right, I missed that bit. Running `find / -type d -inum 2` on my machine actually did produce results in `/dev`, `/run`, and `/net`. One option would be to ignore those three directories, but that feels like a hack.

What about if it checked the inode first, and then only ran `getmntent` if the inode matches?

daviessm:
I think that's what I'd do, yeah. However it's beyond my Rust skills at the moment so I'll do it when I get around to it later unless you want to pick it up first.

I was also thinking about caching all the `getmntent`s somewhere once they've been looked up, because there might be lots of directories within directories, all of which are subvolumes, and that would result in a lot of lookups. Could they be used for anything else (like highlighting mount points themselves...)?

daviessm:
New PR: #941

PR #452, opened 2018-12-08
 Highlight btrfs subvolumes and mount points; output mount point info

daviessm:
Ref: https://github.com/ogham/exa/issues/434

I don't expect you to pull this straight in; I'm not a Rust developer so a lot of this code probably isn't very Rust-y and will need error handling and fixups.

That said, I've tested locally and it does what I think it should: identifies `btrfs` subvolumes and highlights them with the `^` character in the first column (list view) and funky colouring. Mount points are also underlined in all views and in list view include an arrow `->` pointing to the device name and filesystem type. That would probably be better as an option, of course.

The code for looking up mount points was borrowed from https://github.com/acdha/mountstatus/tree/master/src/get_mounts and expanded to get the fs type and device name. It would be nice to also get the mount options because the mount point alone isn't very useful for btrfs at least. Also, I don't have a non-Linux system to test on so `bsd.rs` is almost certainly broken.

There's also the case that a directory could be both a mount point and a subvolume - i.e. a btrfs subvolume is mounted at that point. I'd like to highlight that in the colouring somehow but couldn't think of a good way to do it (background and underline maybe). Likewise, first character in list view would be nice to have different for mount points. Perhaps it would be possible to do something funky with colours.

Comments welcome.


ariasuni:
I was almost sure I left a comment here a few weeks/months ago, but apparently not… Would you be able to update your code so that it can be merged? That sounds like an interesting feature to have, especially since Btrfs is used a lot nowadays.

daviessm:
Yeah, I've been meaning to do this for ages! I'll send a new PR when I've rebased it (or rewritten it with my improved Rust knowledge).

daviessm:
Superseded by #941

PR #457, opened 2018-12-20
Fix sorting by created time

### LINKED ISSUE #325: `ctime` does not mean `created` but `changed`
The `--created` option is a misnomer. The `ctime` tells you the time of the last change to the file status information, e.g. a change to the file permissions.

### LINKED ISSUE #445: exa sort on created time is incorrect
I noticed this code:

https://github.com/ogham/exa/blob/485611e7c97d2043731ae83653a70eee2eb69a4b/src/fs/file.rs#L306-L312

According to Rust documentation, `ctime()` [returns the time of the last status change of the file](https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.ctime).

exa should probably use `accessed()`, `modified()` and `created()` from [`Metadata`](https://doc.rust-lang.org/std/fs/struct.Metadata.html). The only caveat I see is that exa should handle properly the fact that some Unix OS don’t support creation time — namely Linux (how to handle completion files?).

(among Unix OSes, `created()` returns a value for `netbsd`, `freebsd`, `openbsd`, `macos` and `ios` according to [Rust source code](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fs.rs))

ariasuni:
- Use correct function to get `created time` metadata
- Introduce “changed time” to have the former behavior
- Make sure every part of the source code knows about this new flag
- Display error to the user instead of panicking when feature is not supported
- Fix and add some tests

Fix #445.
Fix #325.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ These options are available when running with --long (`-l`):
 - **--time-style**: how to format timestamps
 
 - Valid **--color** options are **always**, **automatic**, and **never**.
-- Valid sort fields are **accessed**, **created**, **extension**, **Extension**, **inode**, **modified**, **name**, **Name**, **size**, **type**, and **none**. Fields starting with a capital letter sort uppercase before lowercase. The modified field has the aliases **date**, **time**, and **newest**, while its reverse has the aliases **age** and **oldest**.
-- Valid time fields are **modified**, **accessed**, and **created**.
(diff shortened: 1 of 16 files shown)


ogham:
This is really good to have done so consistently; thank you. Sorry about the wait.

ariasuni:
Hi! I’m super glad you merged this because I was unsure about some choices I made here. I sure hope this will help people trying to port exa to other platforms, especially Windows, if they want to.

PR #461, opened 2019-01-06
Upgrade git2 and added flag vendored-openssl

cecton:
To make the compilation with x86_64-unknown-linux-musl works I need openssl-sys to be compiled with the feature "vendored" (see sfackler/rust-openssl#980)

The feature flag is present in curl-sys and git2 but only starting from the version 0.8 of git2. So I upgraded git2.

exa can now be compiled with MUSL:

```
cargo build --release --target=x86_64-unknown-linux-musl --features vendored-openssl,git
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,21 +36,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cc"
-version = "1.0.6"
+version = "1.0.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cfg-if"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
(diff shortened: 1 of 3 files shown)


cecton:
Hello @ogham ,

While sorting out my dotfiles I wanted to switch some scripts that compiles in favor of downloading statically linked binaries. I could do it for [bat](https://github.com/sharkdp/bat/releases) but for exa there are only dynamically linked binaries. So I was going to open an issue but I found my PR instead which I completely forgot.

It would be great to merge it and have binaries compiled with musl in the releases when you have time.

Thanks a lot for your work, I love it

cuviper:
FWIW, git2 0.9 is available now.

ogham:
Hi,

Sorry for the really slow response to this, I was worried about making changes to exa's git functionality. The problem that happened last time doesn't seem to be happening anymore though so I have now merged this. Thanks 👍

PR #478, opened 2019-02-12
Added the color support (Orange) for .json files

almir-campos:
* Important: I didn't reformatted the code to not make more changes than required by the feature.


ogham:
Hi,

Sorry, but I don’t want to include default styles for plain text formats like this (or for things like `.xml` or even `.rs` files).

If you want to customise exa on your own machine, you can set the `EXA_COLORS` environment variable to the string you have in styles.rs, `*.json=38;5;202`.

PR #481, opened 2019-02-25
Improve zsh completion

mgkurtz:
Make the `zsh` completion support option stacking as in `exa -al` and delimiting with `--` as  in `exa -- -a`. Also made the completion allow separating arguments for options with ` ` or `=` as in `exa --sort=name`.

For convenience added a comment to the beginning of the file, saying where to put it.


ogham:
Thanks!

Issue #506, opened 2019-06-05
Times headed by zero (such as 00:11) are incorrectly printed (they show up as 0:11)


jorgedfbranco:
Hi,

When running `exa --tree --long` I get:

![image](https://user-images.githubusercontent.com/677887/58994187-f17dd580-87ef-11e9-8872-19a19f2942e0.png)

while `ls -l` will show 00:11.

Thanks,
Jorge

ogham:
A fix has been merged and will go out with the next release.

PR #509, opened 2019-06-11
Convert default formatting so single-digit hour times are 0-padded

emgelb:
Adjusts formatting of times to better resemble standard ls, by adding zero-padding for hours digit (as suggested by [Issue 506](https://github.com/ogham/exa/issues/506).

Example output (before):
```
drwxr-xr-x    - root 10 Jun  2:00 timeshift
```

Example output (after):

```
drwxr-xr-x    - root 10 Jun 02:00 timeshift
```


ogham:
Thanks! 👍 

I think I'll have to find some way to put the locale back into the recent date format, but I'm not sure how to do that yet (and all the tests still pass)

PR #516, opened 2019-07-02
Include symlinks to dirs when grouping dirs

### LINKED ISSUE #471: [Feature Request] Put symlinks to directories first when using `--group-directories-first`
Currently, when using `--group-directories-first` directories and symlinks to directories are not put together, which would be useful considering they're treated mostly the same way (they're both for cd'ing into and organizing files).

VichoReyes:
Closes #471.
When using the option --group-directories-first, symlinks to directories are now first too.


ogham:
Thanks! 👍

PR #533, opened 2019-07-17
Hide "i" in binary prefixes

### LINKED ISSUE #513: Hide "i" in binary prefixes
With the `--binary` (`-b`) option, exa displays binary prefixes with an "i", e.g. `Ki`, `Mi`, unlike ls. This adds clutter and often makes the output 1 column wider without any benefit; the user knows when they have specified `--binary` and thus would know to read them as powers of 1024. It would be nice if exa didn't display the "i" (or at least if there were an option to hide it).

j-tai:
Fixes #513


ogham:
Hi,

Thanks for your contribution. I'm sorry but I'm not going to merge this in, because I don't think it's a good idea to use the shorter prefixes _by default_. A user could be new, and it's easier to see what the `--binary` flag does if it changes the unit as well, or they could have typoed the `-B` flag and introduced a change unexpectedly.

Would you mind if I placed this behaviour behind an environment variable? You will still get credit for the commit.

j-tai:
Go ahead!

PR #539, opened 2019-07-20
Fix readme list rendering

cjbassi:
List wasn't rendered properly since markdown is picky about empty lines in lists


ogham:
Hi, thanks for your contribution. The problem with the README confused me for a long time! However, I plan on fixing it in a different way, so I'm not going to merge this.

Issue #556, opened 2019-07-31
Error thrown if file time metadata is invalid.


BrennanMcDonald:
## Problem
When using a command that reports file time metadata (e.g. `exa -l`) in a directory with files or subdirectories that have invalid (before `SystemTime::UNIX_EPOCH`) we get an error:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SystemTimeError(6795364578.871345152s)', src/libcore/result.rs:999:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
[1]    84324 abort      exa -l
```

## Expected Behaviour
It is expected that when listing a file with an inaccurate time we would see one of the following:
 - Epoch Time
 - Current Time
 - Null/empty string

## Root Cause
In [/src/fs/file.rs](https://github.com/ogham/exa/blob/89c861f24d18650fc3cef50900e3341b5b038ccf/src/fs/file.rs#L329) there is no check to ensure that `self.metadata.modified()` comes after `UNIX_EPOCH`
```
pub fn modified_time(&self) -> Duration {
    self.metadata.modified().unwrap().duration_since(UNIX_EPOCH).unwrap()
}
```
## Proposed Solution
Adding a check in `file.rs` to ensure that the file's time (modified, accessed, created) falls after `UNIX_EPOCH`

```
pub fn accessed_time(&self) -> Duration {
    if self.metadata.modified().unwrap() < UNIX_EPOCH{
        return SystemTime::now().duration_since(UNIX_EPOCH).unwrap()
    } else {
        return self.metadata.created().unwrap().duration_since(UNIX_EPOCH).unwrap()
    }
}
```

ogham:
The fix for this (#670) has been merged and will go out with the next release.

olbrew:
> The fix for this (#670) has been merged and will go out with the next release.

Do you have an estimation of when this next release would be? I've been seeing this error on a static website directory for over a year. A release to fix this would be really nice! :)

ariasuni:
Dozens of people already asked for a new release, please don’t.

In the meantime, you can use `touch` on the offending files (or build exa from source).

PR #557, opened 2019-07-31
Added checks for modified/accessed/created metadata

BrennanMcDonald:
Added checks to `file.rs` to ensure that file's metadata exists after `UNIX_EPOCH`.
If the file was accessed/modified/created after UNIX_EPOCH, the current day is displayed.

See #556

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -5,7 +5,7 @@ use std::io::Error as IOError;
 use std::io::Result as IOResult;
 use std::os::unix::fs::{MetadataExt, PermissionsExt, FileTypeExt};
 use std::path::{Path, PathBuf};
-use std::time::{UNIX_EPOCH, Duration};
+use std::time::{SystemTime, UNIX_EPOCH, Duration};
 
 use fs::dir::Dir;
 use fs::fields as f;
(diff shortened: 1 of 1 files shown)


src/fs/file.rs
```diff
@@ -337,13 +342,23 @@ impl<'dir> File<'dir> {
     }
 
     /// This file’s last accessed timestamp.
+    /// If the file's time is invalid, assume it was accessed today
     pub fn accessed_time(&self) -> Duration {
-        self.metadata.accessed().unwrap().duration_since(UNIX_EPOCH).unwrap()
+        if self.metadata.modified().unwrap() < UNIX_EPOCH{
```
sschlesier: Is this a copy paste error converting `accessed` to `modified`?

src/fs/file.rs
```diff
@@ -337,13 +342,23 @@ impl<'dir> File<'dir> {
...
+        } else {
+            return self.metadata.created().unwrap().duration_since(UNIX_EPOCH).unwrap()
+        }
     }
 
     /// This file’s created timestamp.
+    /// If the file's time is invalid, assume it was created today
     pub fn created_time(&self) -> Duration {
-        self.metadata.created().unwrap().duration_since(UNIX_EPOCH).unwrap()
+        if self.metadata.modified().unwrap() < UNIX_EPOCH {
```
sschlesier: Is this a copy paste error converting `created` to `modified`?

BrennanMcDonald: Fixed in 215b779d3584d6bc7fb619ef05274807e7de85f8

src/fs/file.rs
BrennanMcDonald: Fixed in 215b779d3584d6bc7fb619ef05274807e7de85f8

BrennanMcDonald:
Good eye! I've fixed this copy-paste issue. Thanks.

ariasuni:
Use today’s time just hide incorrect values by displaying another incorrect values. I think you should return `UNIX_EPOCH` instead:

- It’s much easier to spot that something’s weird if files are from 1970.
- It will look like files are updated at each exa invocation, which is confusing.

Otherwise I think that the idea is good.

ogham:
Thanks for this!

I agree that using the epoch is better than the current time. It would also make it easier to test. I'd like to have it log a warning if those are enabled, just so it's not completely silent.

PR #577, opened 2019-08-29
Add support for suppressing table columns

### LINKED ISSUE #226: Hide user with --long
I'd like to hide the `user` column with `--long` enabled, but there does not seem to be a way to do that?

0rvar:
Makes it possible to get sweet lists like this:

Screenshot 2019-08-29 at 14 35 14

Resolves #226

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ These options are available when running with --long (`-l`):
 - **-@**, **--extended**: list each file's extended attributes and sizes
 - **--git**: list each file's Git status, if tracked or ignored
 - **--time-style**: how to format timestamps
+- **--no-permissions**: suppress the permissions field
+- **--no-filesize**: suppress the filesize field
+- **--no-user**: suppress the user field
+- **--no-time**: suppress the time field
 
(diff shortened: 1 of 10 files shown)


thomasf:
This patch does not update `exa --help`

antonmedv:
This is be cool to see.

0rvar:
> This patch does not update `exa --help`, shell completion scripts or man page

Thanks for pointing this out.
* Updated --help, covered by xtests
* Updated man page and verified with `nroff -man contrib/man/exa.1`
* Updated shell completions, verified the Fish completions but could not get bash/zsh completions working as I don't use those shells enough to know how to drop in temporary completions.

@ogham Is this something you want merged at all?

ogham:
Thanks for this! 👍 

I expect that exa should really have generic arguments for all its parameters, but this is a great start.

learn2reid:
not working

0rvar:
> not working

Thank you for your feedback.

Let's lock this PR against further comments

PR #579, opened 2019-09-10
Remove unused code in File#get_source_files

ericbn:
as the "class", "elc", "hi", "o", "pyc" extensions are first tested in FileExtensions#is_compiled, so removed code is redundant and neved called.


ericbn:
A further step to refactor the code could be moving the File#get_source_files function into FileExtensions, with private visibility, as it's only used by FileExtensions#is compiled. Both functions are closely related.

ogham:
Thanks!

Issue #581, opened 2019-09-14
Tree format is slightly misaligned?


LunarLambda:
I'm not sure if this was deliberate or not, but
The original `tree` program formats the tree as such:

```
.
└── luna
    └── bin
```

`exa -T`, however, prints the branches one space to the left instead:

```
.
└── luna
   └── bin
```

Which makes it look slightly disjointed.

A different "solution" would be adding another branching character:

```
.
└──┬ luna
   └── bin
```

PR #588, opened 2019-10-01
original tree style spacing in trees

brown121407:
Lazy fix for #581


ericbn:
Maybe you could remove one horizontal line, so instead of originally having:
```
.
├── tasks
├── vms
│  └── 0
│     ├── data
│     │  └── Docker.raw
```
we would have:
```
.
├─ tasks
├─ vms
│  └─ 0
│     ├─ data
│     │  └─ Docker.raw
```

ogham:
Thanks for your contribution. In this case, I think I am going to be changing the way the tree is printed in the near future to make it customisable, so I won't merge this.

PR #591, opened 2019-10-03
Use statx to get creation time of files on Linux

ariasuni:
I used the crate [statx-sys](https://crates.io/crates/statx-sys) to get the creation time of files under Linux >= 4.11, as Rust stdlib will not have statx support before some time (since it involves a breaking change).

Notable changes:

- Updated the error message with “(needs Linux >= 4.11)”
- Removed a tests that could either pass or fail if Linux >= 4.11 or not

I’m waiting for [this PR for statx-sys](https://github.com/uHOOCCOOHu/statx-sys/pull/1) to be accepted to be able to use the correct field name right away (see PR for details).

Edit: using libc and bumping its version instead.


ariasuni:
Never mind, a workaround has finally been found to use `statx` if available under Linux, so no modification of the current exa code is necessary to make it work, we just have to wait for a new Rust release:
https://github.com/rust-lang/rust/pull/65094#event-2727409141

Thank you @oxalica :)

PR #598, opened 2019-10-12
Optimize file table creation, fixes #484 as a side-effect

### LINKED ISSUE #484: `-s none` yields random order
When I run `exa -lds none file1 file2 file3`, I believed it would print the info for file1, file2 and file3 in this order. However, each time I run this same command, I get a different random order. I understand that
this is problably the intended behavior: concurrency for fast results, and assuming that `-s none` means that the user doesn't care at all about the order. But I think that there should be at least another option for "please just don't change the order of the files given in the command line".

FliegendeWurst:


ogham:
Thank you! I really like the way you solved the problem.

Issue #425, opened 2018-10-11
Configurable colors for --color-scale


pixelastic:
Hello there,

I really like the idea you implemented of coloring the filesize differently based on its size using `--color-scale`. Unfortunately, the hardcoded colors you chose don't work well with my current configuration and I'm looking for a way to configure them.

Would it be possible to add new entries to `EXA_COLORS` to configure them?

Thanks for your time and energy spent in this project

rugk:
https://github.com/ogham/exa/issues/363 may solve the general issue about hardcoded colors.

novoid:
I do have a similar issue with not being able to read executable items using exa v0.8.0 and a dark theme: https://ibb.co/SwGDTKD

ogham:
@novoid That particular shade of blue is _extremely_ annoying. exa has to choose between dropping a colour from its already-very-limited palette (it only has 16 base colours to play with), or being unreadable for people who don't customise their terminal theme. I've chosen the second option because more and more terminals are providing a non-unreadable default theme these days. 

I've tried to justify my decision in this post: https://bsago.me/blue

ogham:
#607 should makes these configurable.

Documentation for how to customise the UI hasn't landed yet, but the PR has the details.

PR #607, opened 2019-11-17
Allow customizing size scale colours.

de-vri-es:
This PR allows the colours used for file sizes and their units to be configured using `EXA_COLORS`. It adds 10 new configuration keys for that.

The PR also adjust the selection of the colour depending on the scale used. With decimal prefixes colours change on powers of 1000. With binary or no prefixes colours change on powers of 1024.

This is a different approach to solving #425. It's orthogonal to #370. That PR changes the built-in colors, this PR makes them user configurable.

I personally have one big worry with this PR: the number of two letter combinations for configuration keys isn't *that* big, and this adds 10 of them. There's more things in exa that I'd like to be user configurable that currently aren't, but finding logical two-letter codes for them may become hard.

I was considering simply using a different environment variable (like `EXA_SIZE_COLORS`), but that seemed like something to discuss first. Advantage is that it can use the same parse code and doesn't need filesystem access like a configuration file would. And it does add a namespace to avoid conflicts with future options. The config keys could even be longer than two letters in a different environment variable.

Another option could be to come up with a syntax that allows for configuration keys longer than 2 letters in `EXA_COLORS`, but that seems more difficult.


de-vri-es:
Here's a picture of what can look like the way I would like to configure it:

exa-configure-scale-colors

ogham:
Thank you for this! It's in line with what I was looking for with this feature. I want to make a lot more of the interface configurable too, so one way or another, these styles would have ended up with two-letter codes.

> I personally have one big worry with this PR: the number of two letter combinations for configuration keys isn't that big, and this adds 10 of them. There's more things in exa that I'd like to be user configurable that currently aren't, but finding logical two-letter codes for them may become hard.

I agree. My game plan for the `EXA_COLORS` environment variable is to keep adding codes for UI elements until the situation becomes unwieldy, at which point it will be time to switch to something else. There has to be something better than one long string of numbers and symbols, but until more things are made configurable, I don't want to switch to something that might not end up being the right way. Having multiple environment variables is an interesting idea!

de-vri-es:
Cool, thanks for the review and merge :)

ariasuni:
It fixes #607 I guess.

PR #616, opened 2019-12-20
Fix handling of potentially unsupported time metadata

ariasuni:
The problem:

- We can’t use creation time on some configurations because it checks on the temp directory that can be on `tmpfs`, which doesn’t support creation time
- It will panic if the check is successful but tries to display a file on a filesystem that doesn’t support creation time.

So basically I discovered that my first attempt to tackle this issue was… not good at all. And this one has the benefit to be simpler.

---

- Checking on a directory doesn’t tell us if supported elsewhere (some filesystems, like tmpfs, don’t support created time)
- We want to be able to display a column even if some subfiles or subdirectories don’t support it

So now if unsupported a time of zero since Epoch is used, and displayed as `-`.

---

A duration of zero since EPOCH is almost always a bug or a special case, so I think it is not a problem that it is sorted first when a sort on e.g. creation time is used.

Sample output:

```
     Running `target/debug/exa -l --created --sort=created /`
dr-xr-xr-x - root -              sys
drwxr-xr-x - root -              run
drwxr-xr-x - root -              boot
drwxr-xr-x - root -              dev
dr-xr-xr-x - root -              proc
drwxrwxrwt - root -              tmp
drwx------ - root  2 juil. 21:51 lost+found
drwxr-xr-x - root  2 juil. 21:54 var
drwxr-xr-x - root  2 juil. 21:54 etc
drwxr-xr-x - root  2 juil. 22:47 usr
drwxr-xr-x - root  2 juil. 22:47 home
drwxr-x--- - root  2 juil. 22:47 root
drwxr-xr-x - root  2 juil. 22:47 opt
drwxr-xr-x - root  2 juil. 22:47 mnt
drwxr-xr-x - root  2 juil. 22:47 srv
lrwxrwxrwx 7 root 19 nov.   1:47 bin -> usr/bin
lrwxrwxrwx 7 root 19 nov.   1:47 lib64 -> usr/lib
lrwxrwxrwx 7 root 19 nov.   1:47 sbin -> usr/bin
lrwxrwxrwx 7 root 19 nov.   1:47 lib -> usr/lib
```

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -328,7 +328,10 @@ impl<'dir> File<'dir> {
 
     /// This file’s last modified timestamp.
     pub fn modified_time(&self) -> Duration {
-        self.metadata.modified().unwrap().duration_since(UNIX_EPOCH).unwrap()
+        match self.metadata.modified() {
+            Ok(system_time) => system_time.duration_since(UNIX_EPOCH).unwrap(),
+            Err(_) => Duration::new(0, 0),
+        }
     }
 
     /// This file’s last changed timestamp.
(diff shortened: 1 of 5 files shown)


ogham:
Hey,

I'm glad you made a PR for this — thanks! I was reading through the code just now and saw the temporary-directory code, and it took me a while to see how it connected with the rest of exa. I agree with your changes and think your new way is an improvement.

PR #619, opened 2019-12-31
Better icons spacing

aeonzh:
Hi, 
I just added a whitespace, in order to make the output a bit nicer (at least to me).
Thought it might be worth a PR in case someone else is interested.
Here are the preview comparisons:

macOS Terminal:
terminal

iTerm2:
iterm

kitty:
kitty
Apparently kitty tries to center non-ASCII glyphs in two cells if followed by two whitespaces, this leads to a bit weird aligning. (I believe I came across an issue regarding this)

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -38,7 +38,7 @@ pub fn painted_icon(file: &File, style: &FileStyle) -> String {
                     c.paint(file_icon).to_string() 
                 }
             });
-    format!("{} ", painted)
+    format!("{}  ", painted)
 }
 
 fn icon(file: &File) -> char {


ogham:
I agree, this does look nicer. Thanks!

horosgrisa:
@ogham Now look very ugly
image

ogham:
Oh... damn

@aeonzh, @horosgrisa, can I ask what fonts you are both using, and whether they were modified to contain icons?

horosgrisa:
@ogham I use JetBrains Mono, but I think that icons come from the fallback font.
See old exa:
image

And new:
image

horosgrisa:
@ogham I think that @aeonzh must use a better font, who draws these symbols as single-width characters

aeonzh:
Oh I see...  
I am using [Fantasque Sans Mono Nerd Font](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/FantasqueSansMono) with the double-width glyphs (it has the single-width option too, if I remember correctly) because in my opinion those were nicer (especially, for example, on icons like Markdown or Rust).
I am not sure if there is a way to detect if the Devicons are drawn as double-width glyph or single ones.

Again, it was my personal taste so I can just patch it on my own machine while you can revert the change I made.

CallumHoward:
This change seems to have not been taken into account when calculating the grid-view spacing.
Screen Shot 2020-07-25 at 11 52 50 am

ogham:
@CallumHoward That's just been fixed in f1e3e7c7ff1228f5ffb86eddb125d29a7ee5f871 and will go out in the next release. I _knew_ I saw the issue on GitHub somewhere, but couldn't find it while I was writing the commit message.

CallumHoward:
@ogham much appreciated! Looking forward to trying it out

PR #622, opened 2020-01-13
Don't make assumptions about Metadata based on temp_dir

mrkline:
Previously, exa would attempt to get the metadata for
`std::env::temp_dir`, and if fetching a certain timestamp
(e.g., creation time) failed, it would assume that the OS couldn't
produce that stamp for _any_ file. Requests to display or sort by that
timestamp would misfire the entire program.

This caused exa to make some sweeping claims that aren't true; e.g., that
Linux can't produce file creation times. The real story is that tmpfs
(the filesystem usually used for `/tmp`) can't, but other Linux
filesystems can and do. (The Rust standard library docs even mention
that `Metadata.created()` is backed by `statx()` since Linux 4.11.)

Since exa could be asked to list files from several filesystems,
let's handle metadata on a per-file basis. This patch makes File's
`*_time()` methods return an Option<Duration>
(`changed_time()` always returns None on a non-Unix system),
and prints a hyphen for files without that particular timestamp.
For example:

        $ exa -lhUd --sort=created /tmp/example Cargo.toml Cargo.lock src target

        Permissions Size User    Date Created Name
        .rw-r--r--     0 mrkline -            /tmp/example
        .rw-r--r--   25k mrkline 12 Jan 22:35 Cargo.lock
        .rw-r--r--  1.2k mrkline 12 Jan 22:35 Cargo.toml
        drwxr-xr-x     - mrkline 12 Jan 22:35 src
        drwxr-xr-x     - mrkline 12 Jan 22:46 target

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -1,6 +1,6 @@
 //! Files, and methods and fields to access their metadata.
 
-use std::fs::{self, metadata};
+use std::fs;
 use std::io::Error as IOError;
 use std::io::Result as IOResult;
 use std::os::unix::fs::{MetadataExt, PermissionsExt, FileTypeExt};
@@ -9,8 +9,6 @@ use std::time::{UNIX_EPOCH, Duration};
 
 use fs::dir::Dir;
 use fs::fields as f;
-use options::Misfire;
-
 
 /// A **File** is a wrapper around one of Rust's Path objects, along with
 /// associated data about the file.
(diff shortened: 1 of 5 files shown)


mrkline:
Ah, apparently using `statx()` in the standard library is relatively new: https://github.com/rust-lang/rust/pull/65094

ogham:
Thanks for your contribution, and for pointing me towards `statx` — it's been a while since I looked at the `stat` manual page and that function was new to me.

In this case, I merged another PR that took care of the temp_dir call, so I'm afraid I can't merge this one.

PR #635, opened 2020-02-15
Decorate symlink targets in long view

daviessm:
Use the absolute path when creating a symlink target - this means that symlink target files are coloured/decorated in long view.


ogham:
Sorry, I can't merge this — the displayed path needs to be the exact path that the symlink has. I've used this to diagnose weird edge cases with relative symlink paths, like where it's missing a leading `/` or something.

exa should already be following the symlink and using the file at the other end to colour the last part of the target path.

PR #643, opened 2020-03-29
Fix SystemTimeError when file's time is earlier than UNIX_EPOCH

Aloxaf:
A file with `1970-01-01 00:00:00.000000000 +0800` as created time
will cause SystemTimeError.


victortoso:
Ah, cool. I just did a similar fix. Small note that this patch extends a636d08 for "potentially unsupported time metadata".

ogham:
Thanks for your contribution. I ended up merging the other PR for this issue.

PR #650, opened 2020-04-11
fixed colored icons being cut off

### LINKED ISSUE #648: Colored icons get cut off.
Hello! It's not that big of a problem, but icons tend to get cut off when color gets applied to their corresponding filetypes. I'm attaching a screenshot, take a look at the .rlib file. 

I'm using Iosevka Term here, the non-mono variant. Since the mono variant has smaller glyphs, the problem  disappears when using that instead, so there's that. Anyways, thanks for all your work, and take care!

Screenshot_20200405_155522

bob16795:
Fixes #648

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -25,7 +25,7 @@ impl Icons {
 }
 
 pub fn painted_icon(file: &File, style: &FileStyle) -> String {
-    let file_icon = icon(&file).to_string();
+    let file_icon = format!("{}  ", icon(&file).to_string());
     let painted = style.exts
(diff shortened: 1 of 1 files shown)


kbc8090:
Is there any chance this will be merged soon?

ogham:
Thanks for your contribution. This has come up before (see #619) — I think it comes down to your terminal emulator's settings. What I'm going to do is add a new environment variable (called `EXA_ICON_SPACING` or something) that can toggle it on and off, and add some recommended examples for fonts and terminal settings on the website.

PR #653, opened 2020-04-19
Use git2 instead of parsing .gitignore for --git-ignore

### LINKED ISSUE #636: --git-ignore parses paths beginning with / incorrectly
If `.` is the root of a Git repository, and I want to ignore `./file.txt` but not `./dir/file.txt`, I can add the line `/file.txt` to `.gitignore`. Git supports this, but exa does not.

ariasuni:
Fix #636 

I can confirm that exa wasn’t able to hide a `node_modules` directory with a `.gitignore` with the line `node_modules/`.

Now, instead of using the [glob](https://crates.io/crates/glob) crate to interpret the lines of `.gitignore`, exa uses part of the code that’s used to display the state of files in a repository.


lucasholder:
This is also an issue in Rust projects (`cargo init` creates a `.gitignore` with a `/target` line). Surprised this wasn't caught earlier given it's a Rust project.

ariasuni:
For some reason, it works for `directory_name` but not for `directory_name`. Default `.gitignore` created by cargo use the first syntax, while I tend to use the form with a slash to indicate a folder in my own `.gitignore`.

Anyway, there are others cases where git patterns are different than glob patterns; it’s just much more reliable and simpler to ask git to tell us what it is ignoring.

kisik21:
I think it also doesn't work on `/directory` patterns, which is common to ignore `node_modules` in root of a Node.js project.

ogham:
This is fantastic! Thank you so much.

ariasuni:
Thanks, I’m proud of this code because I remember that it wasn’t an easy change. Glad it’s useful!

Dreomite:
I'm not sure if this is a bug or intended behavior, but `.git` directory is not ignored when using `--git-ignore`.

ariasuni:
Well, the current behavior is what’s documented: “ignore files mentioned in '.gitignore'.”

I guess it would make sense to not show the `.git` directory but people might also find this behavior unexpected. I guess it might add some complexity too, as e.g. if we hide a `.git` directory it has to be the one at the root of the git repository.

If someone uses `--git-ignore` a lot and wants this behavior, they could use an alias and add the option `-I \.git`. So I guess it could be changed if there is a compelling use case but otherwise I’m not sure it’s worth it.

Dreomite:
You're totally right. I checked my other CLI tools (fd, ripgrep, broot) and all of them behave exactly the same. I got confused because all of these tools ignore hidden files by default, and I forgot for a moment that I have my exa aliased with `-a` for several years.

ariasuni:
exa is unmaintained, please use [the active fork eza](https://github.com/eza-community/eza) (which has regular releases). Also see https://github.com/ogham/exa/issues/1243

PR #654, opened 2020-04-19
Clean code a bit and fix typos in code documentation

ariasuni:


build.rs
```diff
@@ -50,9 +50,10 @@ fn write_statics() -> IOResult<()> {
     use std::io::Write;
     use std::path::PathBuf;
 
-    let ver = match is_development_version() {
-        true   => format!("exa v{} ({} built on {})", cargo_version(), git_hash(), build_date()),
-        false  => format!("exa v{}", cargo_version()),
```
ogham, line 55: I actually really like matching on a boolean, but whenever I show code using it to anyone, I get weird looks. So I suppose this is for the best. _wistful sigh_

ogham:
This is really great, thank you.

build.rs
ariasuni: Oh sorry I realize now that it’s rude due to the name of my PR and the absence of description. As you may have seen in the commit messages, it’s a change suggested by clippy that I tend to follow except if it’s really uglier.

ogham: Don't worry, I didn't take it as rude at all.

PR #655, opened 2020-04-21
Use blue bold instead of blue normal

ericbn:
as blue normal can look like the background color in some dark terminal color schemes. See #370

I suggest avoiding the blue normal going forward, even in future exa color schemes.

### DIFF
--- a/src/options/style.rs
+++ b/src/options/style.rs
@@ -422,31 +422,31 @@ mod customs_test {
     }
 
     // LS_COLORS can affect all of these colours:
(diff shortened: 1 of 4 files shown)


ogham:
Thanks for your contribution. I really do think there are so few options for terminal colours that we can't afford to stop using one of them, and I'm still convinced that default terminal colour schemes are slowly changing for the better. So I'm going to keep the two blues.

However, I've read a lot of comments about this, so I feel like I have to do _something_ to support these terminals. My plan is to include built-in colour schemes that you can refer to by name, so instead of having to read about `LS_COLORS` to change the colour, you can just start using `EXA_THEME=bold_blue` (or something similar).

Issue #535, opened 2019-07-18
--icons icon customization


lilyball:
The new `--icons` flag hardcodes the list of known icons. There should be a means by which I can teach it about new icons to use, or override the existing ones.

I'd also suggest tweaking the default icon from `'\u{f15b}'` (filled document icon) to `'\u{f016}'` (outline document icon), as the filled style seems more emphatic than most of the other icons, and in particular compared to the folder icon. The default icon shouldn't overshadow the icons for more specific filetypes.

lilyball:
Also looking at the list right now, it's got definitions for `.apple` and `.shell`, which don't make much sense (both extensions are apparently associated with the Apple II operating system, which is extremely obsolete). And there's one for `.android` which isn't associated with any known file type at all.

The `.shell` one should probably match file extensions for known shells, such as `.fish`, `.sh`, `.bash`, `.zsh`, etc. The `.apple` one could match stuff like `.xib`, `.nib`, `.xcodeproj`, and `.xcworkspace`, if there's nothing better (nerdfonts.com is nonresponsive right now so I can't check if there's an icon better suited for Xcode-related extensions). I have no idea what extensions would be appropriate to use for `.android`.

mimadrid:
Also a lot of icons are missing compared for example with lsd see https://github.com/Peltoche/lsd/blob/master/src/icon.rs
And many special files like .vimrc, .zshrc,  *.lock are included too in lsd

danielbayley:
> I'd also suggest tweaking the default icon from '\u{f15b}' (filled document icon) to '\u{f016}' (outline document icon)

Agreed.

ddelange:
partial duplicate: #665

obvionaoe:
Where do the icon codes come from? I'd like to make some PRs to incorporate more icons such as the .tf file extension for [Terraform](https://www.terraform.io/) files

ddelange:
@obvionaoe in a [search](https://github.com/ryanoasis/nerd-fonts/search?q=fcd2) of the nerd-fonts repo this file is the only hit: https://github.com/ryanoasis/nerd-fonts/blob/v2.1.0/css/nerd-fonts-generated.css

pgierz:
Has anyone had any experience here? I also would like to add (or override) some of the icons that show up for specific file types.

eggbean:
Read the top of the repo README.md.

Ask at eza

https://github.com/eza-community/eza

ariasuni:
[...]

PR #664, opened 2020-05-03
Improve default file icon

danielbayley:
As mentioned in #535…


ogham:
Thanks! After comparing the icons, I like the outline icon more too.

PR #666, opened 2020-05-14
Show hidden files in Grey and with a different icon.

anxtfux:
Show hidden files in Grey and with a different icon.


ogham:
Thanks for your contribution. For now I'm going to keep the icons' colours matching their file names.

Issue #658, opened 2020-04-26
exa -l fails because of file date


lccro:
```shell
$ RUST_BACKTRACE=full exa -l
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SystemTimeError(6795364578.871345152s)', src/libcore/result.rs:1084:5
stack backtrace:
   0:        0x10ab92ced - std::panicking::default_hook::{{closure}}::hffc69ca72fa98b80
   1:        0x10ab927bc - std::panicking::rust_panic_with_hook::h058ef423beec40ce
   2:        0x10ab92f7d - std::panicking::continue_panic_fmt::h0bb2cdb45a553955
   3:        0x10ab93249 - rust_begin_unwind
   4:        0x10ac24dfa - core::panicking::panic_fmt::h2413d97d26203710
   5:        0x10ac25399 - core::result::unwrap_failed::hef33208e3f23e07a
   6:        0x10ab586bd - exa::output::table::Table::display::h7d61868c2e1aa281
   7:        0x10ab6b955 - <F as scoped_threadpool::FnBox>::call_box::hfeff2cdcb76e7a57
   8:        0x10ab83d6e - std::sys_common::backtrace::__rust_begin_short_backtrace::hf388fc36537a8fe9
   9:        0x10ab8489e - core::ops::function::FnOnce::call_once{{vtable.shim}}::hd49318dcebc5ef76
  10:        0x10ab9d13d - <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once::hf778f932793a5a82
  11:        0x10ab9d767 - std::sys::unix::thread::Thread::new::thread_start::ha4ccb7099fa6037f
  12:     0x7fff6f001109 - _pthread_start
[1]    35138 abort      RUST_BACKTRACE=full exa -l
```

Hope this helps:

```shell
$ stat public
  File: public/
  Size: 640             Blocks: 0          IO Block: 4096   directory
Device: 1000004h/16777220d      Inode: 34369309    Links: 20
Access: (0755/drwxr-xr-x)  Uid: (  501/     leo)   Gid: (   20/   staff)
Access: 1754-08-30 19:37:14.()/-,+/+. -0306
Modify: 1754-08-30 19:37:14.()/-,+/+. -0306
Change: 2020-02-29 20:26:46.219092665 -0300
 Birth: 2020-02-29 20:26:45.878952429 -0300
```

I have no idea how it got that date. After touching it solved the issue.

[...]

PR #670, opened 2020-05-18
Handle timestamps before UNIX_EPOCH (#658)

Freaky:
Instead of returning a Duration since the epoch from file metadata,
which cannot represent times before it, return the SystemTime directly.

Move conversion closer to where it's needed, and perform it infallibly.

---

In contrast to #643, this does not clamp files with pre-epoch times to the epoch:

```
.rw-r--r--    0 freaky  1 Jan  1964 hmm
```

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -325,37 +325,36 @@ impl<'dir> File<'dir> {
         }
     }
 
-    /// This file’s last modified timestamp.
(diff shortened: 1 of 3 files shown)


Freaky:
Well, sort of.

noqqe:
Thanks for doing that. 

Built it locally on my machine and works like a charm

ogham:
Thank you! This is great.

PR #690, opened 2020-06-21
Add immediate file types for PHP ecosystem

schlessera:
Added the following imemdiate file types for PHP:
- `composer.json` for [Composer](https://getcomposer.org/), the defactor standard package manager for PHP;
- `RoboFile.php`, for [Robo](https://robo.li/), the PHP task runner.

Note: [Phing]() uses `build.xml`, which was already included in the list.


ogham:
Thanks!

PR #695, opened 2020-07-12
Fix gitignore option when invert mark exists after asterisk

ryuta69:
## Bug

- `--git-ignore` option ignore the `!aaa` after `*` in .gitignore.

# Screenshots

ss

# Why this happens

https://github.com/ogham/exa/blob/master/src/fs/filter.rs#L349

If the Pattern of asterisk exists in .gitignore, `fn matches_path` of glob library returns true, so `any` of `fn is_ignored_path` always returns true.

# remark
I can't think up of a case to need to care about relative path. If it's necessary, I will fix.

### DIFF
--- a/src/fs/filter.rs
+++ b/src/fs/filter.rs
@@ -346,7 +346,19 @@ impl IgnorePatterns {
 
     /// Test whether the given file should be hidden from the results.
     pub fn is_ignored_path(&self, file: &Path) -> bool {
-        self.patterns.iter().any(|p| p.matches_path(file))
+        let iter = self.patterns.iter();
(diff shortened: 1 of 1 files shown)


ogham:
Hi,

We recently merged in (#653) a change to the way git-ignoring works. It now uses the Git library to do it, rather than trying to roll our own.

Is it possible for you to build exa locally and see if this is still an issue in the development version? I tried your use case (+1 for `bat` btw) and it seems to work now.

ryuta69:
@ogham 
I confirmed that it fixes.
I'll close this.

ogham:
Excellent, thank you.

PR #704, opened 2020-07-24
Add --octal-permissions argument

### LINKED ISSUE #316: Feature request: show file permissions numerically
Adding the chmod number to the detailed info list would be nice. Same as `stat --format '%a' <file>` can output

LarsHaalck:
Using --octal_permissions will insert another column before the existing
permissions where permissions are encoded using octal values as
requested in #316

I tried to stay as close to existing code as possible but as I'm relatively new to Rust I'm open to feedback. 

(closes #316)

### DIFF
--- a/src/fs/fields.rs
+++ b/src/fs/fields.rs
@@ -87,6 +87,11 @@ pub struct PermissionsPlus {
 }
 
 
+/// The permissions encoded as octal values
+pub struct OctalPermissions {
+    pub permissions: Permissions,
+}
+
 /// A file’s number of hard links on the filesystem.
 ///
 /// Under Unix, a file can exist on the filesystem only once but appear in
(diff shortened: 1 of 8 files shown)


ogham:
Thanks! Your Rust code is good.

TonioGela:
Thanks @LarsHaalck!

IVIJL:
Tried to use it in exa but getting `exa: Unknown argument --octal_permissions`

informatik01:
> Using --octal_permissions will insert another column before the existing permissions where permissions are encoded using octal values as requested in #316
> 
> I tried to stay as close to existing code as possible but as I'm relatively new to Rust I'm open to feedback.
> 
> (closes #316)

1) There is a typo in the above message: the correct form is `--octal-permissions`, i.e. with the **dash** between the words, not the underscore. I initially copy-pasted this incorrect variant and was disappointed that it doesn't work ("Unknown argument --octal_permissions"). Luckily I eventually found out the reason and got the desired output.

2) It would also be really nice to list this `--octal-permissions` option in the official documentation (man exa). For some reason it is not mentioned there (using the latest as of now version of exa `v0.10.1`).

PR #708, opened 2020-08-02
Use Aug for August

xguerin:


kaushalmodi:
I was bugged by how August messed up the alignment

image

And thankfully I found this :)

PR #709, opened 2020-08-02
Add the documents extension for Apple file formats

poliorcetics:
This adds some extension for document file formats available on Apple's platforms.


ogham:
Thanks!

PR #717, opened 2020-08-14
Update File Type Icons

JavaCafe01:
Added icon for filetype `.sh`.


ogham:
Thanks! There was already an icon for `.sh` down below, but I think I like the new icon better, so I changed them to match.

Issue #699, opened 2020-07-18
Panic on file with pre-Epoch timestamp


alerque:
```
$ touch -t 197001010000 foo
$ ls -l foo
-rw-r--r-- 1 caleb caleb 0 Jan  1  1970 foo
$ exa -l foo
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SystemTimeError(7140s)', src/fs/file.rs:331:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[1]    409858 abort (core dumped)  exa -l foo
```

Note this is a failed attempt at setting an Epoch time stamp, Epoch is actually 2 am, not 0. GNU's `touch` can set this correctly with `touch -d @0`, but BSD's requires the long form format.

Still `exa` probably shouldn't panic in this case.

ogham:
The fix for this has been merged (#670) and will go out with the next release.

PR #721, opened 2020-08-24
Correct panic resulting from pre-epoch file times

Sam-Hart:
Fix for issue https://github.com/ogham/exa/issues/699 where specifying the -l flag would produce a panic if a file had a modified date from before the unix epoch.

This fixes the issue by creating the dates to display by subtracting the elapsed time since the reported event from the current time instead of producing the datetime from the duration since the unix epoch.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 use std::time::{UNIX_EPOCH, Duration};
 
 use log::{debug, error};
-
+use datetime;
 use crate::fs::dir::Dir;
 use crate::fs::fields as f;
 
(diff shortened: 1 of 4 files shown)


Freaky:
The Unix epoch is at least a fixed point in time.  This approach converts from SystemTime to Duration and then to LocalDateTime each time using a brand new Instant, so all timestamps will be temporally smeared based on runtime.

This approach also retains the same problem going from the epoch has, except now it breaks or panics on timestamps in the future:

```
❯ exa -l futur
.rw-r--r-- 0 freaky  1 Jan  3000 futur

❯ target/debug/exa -l futur
.rw-r--r-- 0 freaky - futur

❯ exa --accessed -l futur
.rw-r--r-- 0 freaky  1 Jan  3000 futur

❯ target/debug/exa --accessed -l
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SystemTimeError(30903153086.189168974s)', src/fs/file.rs:376:54
```

Alternate time formatting is also broken:

```
❯ exa --time-style=full-iso -l
.rw-r--r-- 1.9k freaky 2020-09-18 22:20:41.658572524 +0000 build.rs

❯ target/debug/exa --time-style=full-iso -l
.rw-r--r-- 1.9k freaky 1970-01-01 16:51:51.757415542 +0000 build.rs

❯ target/debug/exa --time-style=full-iso -l
.rw-r--r-- 1.9k freaky 1970-01-01 16:51:52.596364124 +0000 build.rs
```

`exa` here is a custom build that includes #670, which I hope is a complete fix for this issue.

ogham:
Thanks for your PR. I've merged in another fix for this issue, so I'm not going to merge this one.

Issue #712, opened 2020-08-05
Unable to determine time zone: Is a directory (os error 21)


Adventure4Life:
My localtime is linked to directory... it is not a file.. this is as I am using a lnux called Bedrock.

Is there anyway I can define the timezone with a --command I can add to the alias... or something like that? 

I can not seem to relink the file directly to the actual local file due to how bedrock works. So I was wondering if there was a way I can force exa to use a specified path to localtime?

ariasuni:
Do you still have this error with exa 0.10.1?

If yes, I would like to know more about how it works on Linux Bedrock.

Also, you can try to use the TZ environment variable (see #725).

Bobbbay:
> Some software may ignore the TZ environment variable and instead attempt to read /etc/localtime to determine the current timezone. Bedrock Linux will copy /bedrock/etc/localtime to /etc/localtime when enabling a stratum to cover this possibility.

TL;DR Bedrock will apparently set both `TZ` and `/etc/localtime`.

Bobbbay:
Also... @ariasuni I should mention this: I'm on Exa 0.10.1 and still receive this error (on WSL + NixOS). My $TZ is set, although /etc/timezone does not exist (just Nix things).

ariasuni:
Please copy the content of your `TZ` variable and the error message when exa crashes. Right I’ve no idea what Bedrock Linux or NixOS is doing, and no idea where exa fails exactly.

I looked at how exa is handling the `TZ` variable and it’s not handled correctly (to be fair, how it’s supposed to work is documented poorly if at all, I had to look at the glibc source code), so I’m working on that, but I’m not sure it will fix your problem.

Bobbbay:
I cannot vouch for Bedrock, as I do not own a Bedrock system and have only dabbled in contributing in the project, but here's some results from my side:

```
[bobbbay@NotYourLaptop:~]$ exa --version
exa - list files on the command-line
v0.10.1 [+git]
https://the.exa.website/

[bobbbay@NotYourLaptop:~]$ exa
... files go here ...

[bobbbay@NotYourLaptop:~]$ exa --long
Unable to determine time zone: No such file or directory (os error 2)
... files go here ...

[bobbbay@NotYourLaptop:~]$ echo $TZ
America/Toronto

[bobbbay@NotYourLaptop:~]$
```

I *should* note the following:

 - `/etc/timezone` does not exist
[...]

PR #725, opened 2020-08-26
Update determine_time_zone function to check TZ

### LINKED ISSUE #453: Exa does not read the `TZ` environment variable, but it is a standard way to convey timezone data on musl
On musl, the way to set the timezone is by having a TZ environment variable.
See [its documentation](https://www.musl-libc.org/doc/1.0.0/manual.html) for reference.

Currently, when running `exa -l`, the following output appears on such a system:
```
Unable to determine time zone: No such file or directory (os error 2)
```

Note that `date` (both GNU and busybox implementation) output the correct time information with no warnings:
```
Fri Dec 14 00:13:10 EST 2018
```

kbravh:
Instead of defaulting immediately to /etc/filename for the timezone, we can first check whether the TZ environment variable is set. If so, we can pull the corresponding timezone file from /usr/share/zoneinfo. Closes #453.

### DIFF
--- a/src/output/table.rs
+++ b/src/output/table.rs
@@ -1,4 +1,5 @@
 use std::cmp::max;
+use std::env;
 use std::fmt;
 use std::ops::Deref;
 use std::sync::{Mutex, MutexGuard};
@@ -286,7 +287,12 @@ impl Environment {
 }
 
 fn determine_time_zone() -> TZResult<TimeZone> {
-    TimeZone::from_file("/etc/localtime")
+    let tz = env::var("TZ");
+    if tz.is_err() {
+        return TimeZone::from_file("/etc/localtime");
+    } else {
+        return TimeZone::from_file(format!("/usr/share/zoneinfo/{}", tz.unwrap()));
+    }
 }
 
 
--- a/src/output/table.rs
+++ b/src/output/table.rs
@@ -287,11 +287,10 @@ impl Environment {
 }
 
 fn determine_time_zone() -> TZResult<TimeZone> {
-    let tz = env::var("TZ");
-    if tz.is_err() {
-        return TimeZone::from_file("/etc/localtime");
+    if let Ok(file) = env::var("TZ") {
+        TimeZone::from_file(format!("/usr/share/zoneinfo/{}", file))
     } else {
-        return TimeZone::from_file(format!("/usr/share/zoneinfo/{}", tz.unwrap()));
+        TimeZone::from_file("/etc/localtime")
     }
 }
 


src/output/table.rs
```diff
@@ -286,7 +287,12 @@ impl Environment {
 }
 
 fn determine_time_zone() -> TZResult<TimeZone> {
-    TimeZone::from_file("/etc/localtime")
+    let tz = env::var("TZ");
+    if tz.is_err() {
+        return TimeZone::from_file("/etc/localtime");
+    } else {
+        return TimeZone::from_file(format!("/usr/share/zoneinfo/{}", tz.unwrap()));
+    }
```
poliorcetics, line 295:
```suggestion
    if let Ok(file) = env::var("TZ") {
        TimeZone::from_file(format!("/usr/share/zoneinfo/{}", file));
    } else {
        TimeZone::from_file("/etc/localtime");
    }
```

This omits the return keyword and avoids an `unwrap`.

Disclaimer: I have not tested this suggestion.

kbravh:
@poliorcetics Thanks for the feedback! I tried out the suggestion you made; I definitely like deconstructing the timezone from the environment variable instead of unwrapping, so I've added that change. 

The only issue is that we have to use the `return` keyword from inside the if block; we can only omit the `return` if we want to return the last expression in a function, otherwise it returns an empty type `()`. So I've followed the pattern on the [Rust docs function page](https://doc.rust-lang.org/rust-by-example/fn.html) and I'm using `return` for the first case where the TZ variable is set. Otherwise, I've left the final expression as the last statement so it's automatically returned. 

```rust
if let Ok(file) = env::var("TZ") {
        return TimeZone::from_file(format!("/usr/share/zoneinfo/{}", file));
    }
    
TimeZone::from_file("/etc/localtime")
```

poliorcetics:
That's because I put `;` in my suggestion, the fix should be:

```rust
    if let Ok(file) = env::var("TZ") {
        TimeZone::from_file(format!("/usr/share/zoneinfo/{}", file)) // no ;
    } else {
        TimeZone::from_file("/etc/localtime") // no ;
    }
```

kbravh:
@poliorcetics That did the trick! Thanks for the help.

ogham:
Thanks for this!

themadsens:
Just bitten by this .. Out of curisosity: Why not use libc::mktime? Then all of these timezone issues would just disappear!

ariasuni:
@themadsens Well I can’t speak for ogham, but I would gladly use a `libc` function if it would make things easier. Right now, it wouldn’t be very useful, and #712 which is reasonably simple would fix the remaining issues.

It’s certainly an option to drop `zoneinfo_compiled` and `datetime` and just use a few C functions, but I think trying to do things in Rust as much as possible can be useful, either to push for more Rust libraries, or simply to for exa’s code to be an example.

PR #734, opened 2020-09-08
Fix icons for symlinked directories

imsofi:
Furthering #516 of grouping sym-linked directories.
This fixes symlinked folders from getting a generic file icon.


ogham:
Thanks!

PR #735, opened 2020-09-13
add ko to compiled objects

Finermeerkat137:
Since a Linux/BSD kernel object is an object/compiled file, it should be counted as one along with o, pyc, and the others.

### DIFF
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -88,7 +88,7 @@ impl FileExtensions {
     }
 
     fn is_compiled(&self, file: &File) -> bool {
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Sadly this is a dead project, not much action here :(

Finermeerkat137:
That’s true, so I just forked the repo, put all my changes there, and use that instead.

ogham:
Thanks! 🧟

ethancedwards8:
Glad to see you’re back in development @ogham !

ogham:
Thanks :) I've confident this time I'm going to stay.

Finermeerkat137:
Nice :)

PR #743, opened 2020-10-03
Add additional TeX compiled files

jonhoo:
Specifically .fls and .fdb_latexmk, which the popular [`latexmk`](https://www.ctan.org/pkg/latexmk) tool produces.

There are [more](https://tex.stackexchange.com/questions/7770/file-extensions-related-to-latex-etc), but these are particularly common. There are two more that are very common, but I couldn't at first glance see how to add them:

 - `.run.xml` generated by the `biber` bibilography tool
 - `-blx.bib` control file generated by `biblatex`

Might be good to include those too if they're easy to add.

### DIFF
--- a/src/info/sources.rs
+++ b/src/info/sources.rs
@@ -21,6 +21,7 @@ impl<'a> File<'a> {
 
                 "aux" |                                          // TeX: auxiliary file
                 "bbl" |                                          // BibTeX bibliography file
+                "bcf" |                                          // biblatex control file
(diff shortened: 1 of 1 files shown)


ogham:
Thanks!

I agree that we should only have the common ones for now (uncommon ones have a higher chance of being a false positive). It's unfortunate that the compiled file checker is extension-based at the moment; when it gets updated I'll try to add in full filename checking.

PR #752, opened 2020-10-18
Add more immediate files

spk:
Hi, just adding more immediate files let me know if some are not wanted, cheers


ogham:
These all look good to me — the definition of what counts as an "immediate" file has changed a little over the years, but in general, it's a file that, when you're seeing a source code project's folder for a first time, it's the file you interact with next. So READMEs, build tools, command runners, etc. 

Thanks!

PR #759, opened 2020-10-30
feat: New Icons and CLI argument to suppress icons

stuartcarnie:
This PR introduces two improvements:

* a significant number of new icons and includes a comment showing the character to which the extension or file maps to
* A new switch, `--no-icons` to explicitly hide icons, which takes precedence over `--icons`. This switch is useful when using aliases and which to temporarily hide the icons

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -2,6 +2,8 @@ use ansi_term::Style;
 
 use crate::fs::File;
 use crate::info::filetype::FileExtensions;
+use lazy_static::lazy_static;
+use std::collections::HashMap;
 
 
 pub trait FileIcon {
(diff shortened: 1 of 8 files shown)


timbedard:
Oh heck yeah. I was thinking that some of these icons were missing.

src/output/icons.rs
```diff
@@ -42,156 +44,293 @@ pub fn iconify_style<'a>(style: Style) -> Style {
...
+        m.insert(".gitmodules", '\u{f1d3}'); // 
+        m.insert(".rvm", '\u{e21e}'); // 
+        m.insert(".vimrc", '\u{e62b}'); // 
+        m.insert(".vscode", '\u{e70c}'); // 
+        m.insert(".zshrc", '\u{f489}'); // 
+        m.insert("Cargo.lock", '\u{e7a8}'); // 
+        m.insert("bin", '\u{e5fc}'); // 
+        m.insert("config", '\u{e5fc}'); // 
+        m.insert("docker-compose.yml", '\u{f308}'); // 
+        m.insert("dockerfile", '\u{f308}'); // 
```
timbedard, line 69: `Dockerfile` is the conventional file name.
```suggestion
        m.insert("Dockerfile", '\u{f308}'); // 
```

timbedard (changes requested):

ariasuni (changes requested):
I think it’s great. Here what’s missing:

- You should add this flag to the fish completion file too.
- You should add the option to the README.
- You should add the option to the man page.

stuartcarnie:
Hi @ariasuni and thank you for the feedback! I've completed all the suggestions.

ogham:
Thanks!

I don't actually use icons myself (I just never got into using them) which is why they always seem to be lagging behind, missing, etc.

PR #762, opened 2020-12-03
Warm when git feature is disabled instead of ignoring flags

### LINKED ISSUE #761: --git switch not working on Debian-based systems
$ exa --version
exa v0.9.0

$ exa -l --git

On this particular computer, exa is not showing git status information in directories which are being tracked using git. The output is the same as it is without the --git switch.

This is on a particular computer running Pop!_OS 20.10 / bash 5.0.17 (essentially Ubuntu), while other computers are fine, using the same dotfiles.

Is there some reason why the --git switch does not work on this Pop!_OS laptop?

ariasuni:
The flags `--git` and `--git-ignore` are caught early during options parsing, so no more checking for git feature is done elsewhere.

Since `--git-ignore` depends on git too [since recently](https://github.com/ogham/exa/pull/653), remove it from help when git feature is disabled.

Extended attributes now don’t artificially depends on git feature being enabled.

Tell if exa was compiled with git support or not when using --help

Tell in the man that some flags needs exa built with git support 

Fix #761

### DIFF
--- a/man/exa.1.md
+++ b/man/exa.1.md
@@ -101,7 +101,7 @@ Sort fields starting with a capital letter will sort uppercase before lowercase:
 `-I`, `--ignore-glob=GLOBS`
 : Glob patterns, pipe-separated, of files to ignore.
 
-`--git-ignore`
+`--git-ignore` [if exa was built with git support]
 : Do not list files that are ignored by Git.
 
 `--group-directories-first`
@@ -174,7 +174,7 @@ These options are available when running with `--long` (`-l`):
 `-@`, `--extended`
 : List each file’s extended attributes and sizes.
 
-`--git`
+`--git`  [if exa was built with git support]
 : List each file’s Git status, if tracked.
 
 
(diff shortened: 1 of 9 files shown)


ariasuni:
I realize that the way this is done is not ideal at all.

Aborting with an error when git is disabled may prevent sharing alias using git-related features between different machines, so I guess it should be ignored. But then if it’s ignored, we need to take any other occasion to explain the user what’s going on. This is why I believe that instead of  `--help` not displaying anything about `--git` and `--git-ignore`, which may confuse users who know about the git options, it should tell user explicitly that these options exist but are unavailable.

These thoughts are related to the fact that some distributor may not enable git, like it was done for Debian/Ubuntu.

ogham:
Thanks for this!

> This is why I believe that instead of --help not displaying anything about --git and --git-ignore, which may confuse users who know about the git options, it should tell user explicitly that these options exist but are unavailable.

Yeah, this was a big downside of omitting them from the `--help` output. It's not an unknown argument, it's a _known_ one, just a disabled one. The new message is much better.

> Aborting with an error when git is disabled may prevent sharing alias using git-related features between different machines, so I guess it should be ignored. But then if it’s ignored, we need to take any other occasion to explain the user what’s going on.

I think if exa started printing out warnings (`Warning: exa was built without Git support; ignoring --git option`) upon startup, users would want to fix their aliases _anyway_, because the warnings would get annoying.

PR #763, opened 2020-12-05
Support reading arguments from config file and environment variables

### LINKED ISSUE #511: Feature request: configuration files
Having to configure exa using a shell alias is less than convenient. ripgrep allows you to set a RIPGREP_CONFIG_PATH environment variable to point to a file that contains a list of default options, one per line (https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file). I think exa should do this too.

kidonng:
Idea and code adapted from [`bat`](https://github.com/sharkdp/bat).

Basically the same logic as bat:

- Look for `EXA_OPTS` environment variable
- If `EXA_OPTS` is not found, look for `EXA_CONFIG_PATH` and check if it exists
- Load arguments from `EXA_OPTS` or config file, and append the ones defined at runtime

Note that `--generate-config-file` is not implemented but I'd like to if there is any demand.

Doc is not updated for the moment because things may change 😄 

Resolves #511

I know little about Rust so correct me if I made any mistake 😃 (and I learned a lot from making this PR).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,6 +43,7 @@ mod logger;
 mod options;
 mod output;
 mod theme;
+mod config;
 
 
 fn main() {
@@ -50,7 +51,11 @@ fn main() {
 
     logger::configure(env::var_os(vars::EXA_DEBUG));
 
-    let args: Vec<_> = env::args_os().skip(1).collect();
+    let cli_args: Vec<_> = env::args_os().skip(1).collect();
+    let mut args = config::get_args_from_env_var()
+        .unwrap_or_else(config::get_args_from_config_file)
+        .expect("Could not parse configuration file");
+    args.extend(cli_args);
     match Options::parse(args.iter().map(|e| e.as_ref()), &LiveVars) {
         OptionsResult::Ok(options, mut input_paths) => {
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -50,7 +50,13 @@ fn main() {
 
     logger::configure(env::var_os(vars::EXA_DEBUG));
 
-    let args: Vec<_> = env::args_os().skip(1).collect();
+    let cli_args = env::args_os().skip(1);
+    let args = get_args_from_env_var()
```
kidonng, line 54: I'm not sure this is the most performant way, tried to use `append` or `extend` but no luck 😅

jspiro:
Just installed exa and while I'm still unsure about the harder to type name than ls, the lack of config files like ripgrep and bat was a bit too pure. Thank you!

ogham:
Hi, sorry about the delay on this one.

Adding a config file is something I do want to add, but I think exa could do with so many configuration options that merely having the config file be a list of command-line flags is going to be very limiting. I want to keep the option open to have a more structured configuration file.

mfdorst:
I was really hoping some sort of config file was available, as I wanted to do `alias ls='exa -TL3'` but I realized that that would prevent me from specifying a different recursion depth would then be impossible. Is there some workaround for this in the meantime?

PR #781, opened 2021-01-01
fix build.rs with shadow-rs.It's use easy.

baoyachi:


ogham:
Hi,

`shadow` looks interesting, but I would prefer to keep the build script consistent between my projects.

PR #784, opened 2021-01-03
Use same styling for spaces between icon and file name

### LINKED ISSUE #648: Colored icons get cut off.

ariasuni:
This is only visible when selecting both icon and file name, as most terminal emulators then invert background and foreground color.

It also workaround a Konsole bug: https://bugs.kde.org/show_bug.cgi?id=422776

Fix #648

### DIFF
--- a/src/output/file_name.rs
+++ b/src/output/file_name.rs
@@ -132,9 +132,9 @@ impl<'a, 'dir, C: Colours> FileName<'a, 'dir, C> {
             bits.push(style.paint(file_icon));
 
             match spaces_count {
-                1 => bits.push(Style::default().paint(" ")),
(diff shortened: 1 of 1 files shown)


ogham:
Thanks for this. Somehow none of the terminal emulators I test on regularly invert the colours on selection.

Issue #436, opened 2018-11-08
long + grid + header results in an extra column


gimerstedt:
when using long, grid and header options together, an extra header is added:

```
krig@krig ~/tmp/test/ci
$ exa -hlG
Permissions Size User Date Modified Name    Permissions Size User Date Modified Name
.rwxr-xr-x   630 krig  7 Nov 11:43  test
```

i'm not sure if this is intended or not and it's not options that i normally use, still.

```
krig@krig ~/tmp/test/ci
$ exa -v
exa v0.8.0
```

ogham:
Yeah this looks like a bug

ariasuni:
I see that the issue is worse with the `-x` option:
```
exa -hlGx ~/test/
Permissions Size User     Date Modified Name    Permissions Size User Date Modified Name    Permissions Size User Date Modified Name    Permissions Size User Date Modified Name    Permissions Size User Date Modified Name
.rw-r--r--     0 ariasuni 31 Dec 17:38  file
```

PR #790, opened 2021-01-12
Don’t produce extra column sometimes with long grid + header

ariasuni:
The number of necessary columns was computed by producing a grid in different sizes and see if all columns were used. However, if there was two files and we tried to fit them in a 3-column grid, it would produces three headers and all three columns would be used; when trying a 4-column grid, the two supplementary headers would fill the third column and the fourth would be empty; so 3 columns would be used.

Now, when the grid fits into the terminal and the number of columns is exactly the number of files to display, it returns immediately instead of trying bigger grids.

Fixes GH-436.

### DIFF
--- a/xtests/grid-details-view.toml
+++ b/xtests/grid-details-view.toml
@@ -76,3 +76,24 @@ stdout = { file = "outputs/files_paths_long_grid_3col.ansitxt" }
 stderr = { empty = true }
 status = 0
 tags = [ 'env', 'long', 'grid' ]
+
+
+# check if exa is using the minimum number of columns with headers
+
+[[cmd]]
+name = "‘COLUMN=200 exa -lGh’ with one file don’t produce extra columns even if there place for more"
+shell = "exa -lGh /testcases/files/10_bytes"
+environment = { COLUMNS = "200" }
+stdout = { file = "outputs/files_paths_long_grid_header_1file.ansitxt" }
+stderr = { empty = true }
+status = 0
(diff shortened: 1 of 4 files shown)


xtests/grid-details-view.toml
```diff
@@ -76,3 +76,24 @@ stdout = { file = "outputs/files_paths_long_grid_3col.ansitxt" }
 stderr = { empty = true }
 status = 0
 tags = [ 'env', 'long', 'grid' ]
+
+
+# check if exa is using the minimum number of columns with headers
+
+[[cmd]]
+name = "‘COLUMN=200 exa -lGh’ with one file don’t produce extra columns even if there place for more"
```
YounesTea, line 84: A little typo here : `there **is** place for more`

xtests/grid-details-view.toml
```diff
@@ -76,3 +76,24 @@ stdout = { file = "outputs/files_paths_long_grid_3col.ansitxt" }
...
+shell = "exa -lGh /testcases/files/10_bytes"
+environment = { COLUMNS = "200" }
+stdout = { file = "outputs/files_paths_long_grid_header_1file.ansitxt" }
+stderr = { empty = true }
+status = 0
+tags = [ 'env', 'long', 'grid' ]
+
+[[cmd]]
+name = "‘COLUMN=200 exa -lGh’ with several files don’t produce extra columns even if there place for more"
```
YounesTea, line 93: same typo

ogham:
Perfect, thank you again.

PR #793, opened 2021-01-18
Added sudo to installation command for Linux

DesmondWillowbrook:
Don't know about the need for `sudo` for the rest of the operating systems.
Feel free to suggest those as well.


ogham:
Cheers. There should probably be a "you may need sudo" message covering the whole group. (Homebrew for example doesn't use sudo, but almost all package managers do.)

Issue #730, opened 2020-09-02
Feature Request: Show git branch, status for subdirectories that are git repos


yamlCase:
It would be really awesome if we could have a column for git status and branch when running `exa` in a directory containing multiple git repos. Here's an example from a refactored version of `k`:

![image](https://user-images.githubusercontent.com/27447701/92010639-a5318600-ed0f-11ea-9e4a-d6740015f342.png)

alpn:
Came here to suggest the same thing.
@ogham would you be open to such a contribution?

ariasuni:
I’m not sure if it’s very useful to know on which branch a repository is interesting in exa, except maybe if you use [worktrees](https://git-scm.com/docs/git-worktree), but I guess it makes sense.

By displaying a “git status”, you mean displaying if the repository is ahead or behind the upstream branch?

alpn:
While I can't speak for @yamlCase, for *me* the crucial part here is showing whether or not there are uncommitted changes in the repo.

yamlCase:
> While I can't speak for @yamlCase, for *me* the crucial part here is showing whether or not there are uncommitted changes in the repo.

You can speak for me because that's exactly what I would want it for ;)

Also: being a little OCD, I really like to know all my projects are on the proper branch at the day's end.

alpn:
PR #797 is a proposed implementation of this feature

![image](https://user-images.githubusercontent.com/15992847/112681703-76166980-8e45-11eb-9dbe-b650ccfa6d38.png)

I’d love to get your feedback on it @ariasuni @yamlCase

yamlCase:
> I’d love to get your feedback on it @ariasuni @yamlCase 

I really like the way that looks.  From a UX perspective this has my approval. Can't wait to try it out!

Issue #420, opened 2018-10-05
Feature Request: Support --git (or similar) from directory containing multiple repositories.


coreyfarrell:
I'd like the ability to run `exa -lh --git` from a directory containing multiple repositories, but which is not itself a repository.  For a simple example:
```
cd /usr/src
git clone https://github.com/ogham/exa/
touch exa/make-it-dirty
exa -lh --git
```
Currently no git information is displayed since /usr/src is not a git working directory.  Obviously in real life I would run this from within a folder which contains dozens of cloned repos.  The important information I'd want to see for each repository is:
1. do any uncommitted non-gitignore'd changes exist
2. is the current branch a different commit from remote (only if the current branch tracks a remote branch).
3. (optional) displaying the name of currently checked out branch would be nice but not strictly needed.

PR #797, opened 2021-01-31
Add status for subdirectories that are git repositories

alpn:
This PR adds an optional `Repo` column, showing git information about repositories located directly under the currently searched path, as requested in #730 and #420.

image

Currently it piggybacks the `—git` flag (and works in conjunction with it)


ariasuni:
To be honest I don’t really care about this feature so it wasn’t in my priorities, but it looks very small and clean so I think I’d like to merge it.

However:
- What’s X and V? These labels are not intuitive in my opinion
- I wonder if it should be the same as `--git` or split into its own flag

ariasuni:
In fact I think it’s a cool feature if it could works with git submodules. Do you think you could check that?

alpn:
Thanks for the feedback @ariasuni

The X and V labels indicate the status of the the entire repository as either dirty (e.g there are uncommitted changes) or clean (the equivalent of “git status” returning “nothing to commit, working tree clean“) respectively.

Submodules already get detected as repos in their own right, and as such, their status will be shown as well.

YAMLcase:
> How would you suggest to make it more intuitive?

One option: 
Nothing or an underscore/dash for clean, 1-9 for number of changed files or lines.  9 would indicate "9 or more" like some badge systems that stick to a single digit.

alpn:
Thanks @YAMLcase, but honestly I prefer to keep it simple and stick to an explicit binary indicator (i.e, either _clear_ or _dirty_)
However, here’s another option inspired by your comment:

image


Do you find it more intuitive @ariasuni?

I too was debating about adding a separate flag vs piggybacking `--git`, but had eventually decided to go with the latter because:

1. Having used this feature myself for a few months now, I never felt like a special flag was missing.
2. It’s simpler, and requires less changes to the code base.

That being said, I’m open to adding it if you think it’s crucial.

Otherwise, I’ve fixed the conflicts with `master` so it’s should be mergeable now.

adamshand:
I'd love to see this merged!

alpn:
Up to date with upstream again.
@ariasuni please let me know if there’s anything else I could do to get it merged.

src/fs/feature/git.rs
```diff
@@ -334,3 +334,46 @@ fn index_status(status: git2::Status) -> f::GitStatus {
...
+    pub fn from_path(dir : &Path) -> Self{
+
+        let path = &reorient(&dir);
+        let g = git2::Repository::open(path);
+        if let Ok(repo) = g{
+
+            let branch = current_branch(&repo);
+            match repo.statuses(None) {
+                Ok(es) => {
+                    if es.iter().filter(|s| s.status() != git2::Status::IGNORED).count() > 0{
```
bew, line 367:
```suggestion
                    if es.iter().filter(|s| s.status() != git2::Status::IGNORED).any(|| true) {
```
Should be a bit faster on big repos with a lot of changes, since you only want to check if there is more than zero item.

bew (commented):
For the branch name, you're showing only short branch names.
It's not uncommon to work with medium to long branch names like `fix-for-this-long-feature-name`, I fear the result would be a lot of blank spaces..
Would it be possible to add an option to limit the size of the displayed branch name ? (like 10 or 15 chars max)

alpn:
> For the branch name, you're showing only short branch names. It's not uncommon to work with medium to long branch names like `fix-for-this-long-feature-name`, I fear the result would be a lot of blank spaces.. Would it be possible to add an option to limit the size of the displayed branch name ? (like 10 or 15 chars max)

Thanks @bew, good point!
Added “indicative” truncation when branch name exceeds 10 chars, so this:

image

Becomes this:

image

arazabishov:
This is awesome! Thank you for making this change. My 2 cents, assuming that git status is used to get branch/status info. 
- Some git repos are massive and running git status on them can be very slow. It would be great to have a flag or a config option to switch off the clean/dirty column.
- With branch names we have several options that are faster than git status, such as git rev-parse --abbrev-ref HEAD.

ariasuni:
Exa uses libgit2 directly (which is slow anyway…)…

arazabishov:
What is left on this PR to get addressed before it can be merged?

alpn:
Thanks for the feedback @arazabishov

I’ve just added two dedicated flags for this feature:

`--git-repos` - shows both the status, and the branch name (like in the screenshots above)
` --git-repos-no-status` - skips the status lookup altogether and is hence much faster, as expected :)

Running exa with `--git-repos-no-status` returns almost immediately, even for directories with lots of repos (including, e.g [linux](https://github.com/torvalds/linux), and [openbsd](https://github.com/openbsd/src))

alpn:
Fixed some minor conflicts, so it's gtg again.
Would still love to see this one merged in @ariasuni :)

ariasuni:

PR #801, opened 2021-02-19
documentation: add hint how to install exa on Android / Termux

alexanderadam:
Hi Benjamin,

thank you so much for creating `exa` :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,12 @@ On Arch, install the [`exa`](https://www.archlinux.org/packages/community/x86_64
 
     $ pacman -S exa
 
+### Android / Termux
+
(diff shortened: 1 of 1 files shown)


ogham:
Thanks!

I went on the [termux website](https://wiki.termux.com/wiki/Package_Management) and found this:

> We strongly recommend to use a pkg utility instead of apt directly. 

Would it be better to use `pkg` here?

alexanderadam:
> Would it be better to use `pkg` here?

You are absolutely right. Wasn't aware of that. I changed it accordingly :+1:

ogham:
Thanks!

alexanderadam:
I have to thank you! `exa` is _really_ great! :heart:

PR #813, opened 2021-03-19
Add options for -n or --numeric-gid-uid.

### LINKED ISSUE #807: Show numeric UID and GID
Please add a parameter equivalent to ls' `-n, --numeric-uid-gid` parameter that lists the user and group with their numeric IDs instead of the names.

This is especially pretty handy when it comes to debugging file access with docker containers, as user-names in the container might be mapped to different numeric IDs on the host.

b05902132:
Closes #807.

This option is only available in -l.
There's test for rendering, but no test for option parsing.
(I don't understand that section of code.)

Also in `ls` this option implies `-l`, is it worth it to follow behavior of `ls`?

### DIFF
--- a/completions/completions.fish
+++ b/completions/completions.fish
@@ -65,7 +65,7 @@ complete -c exa -s 't' -l 'time'  -x -d "Which timestamp field to list" -a "
     created\t'Display created time'
 "
 complete -c exa -s 'm' -l 'modified'      -d "Use the modified timestamp field"
-complete -c exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
+complete -c exa -s 'n' -l 'numeric'       -d "List numeric user and group IDs."
 complete -c exa        -l 'changed'       -d "Use the changed timestamp field"
 complete -c exa -s 'u' -l 'accessed'      -d "Use the accessed timestamp field"
 complete -c exa -s 'U' -l 'created'       -d "Use the created timestamp field"
--- a/completions/completions.fish
+++ b/completions/completions.fish
@@ -65,7 +65,7 @@ complete -c exa -s 't' -l 'time'  -x -d "Which timestamp field to list" -a "
     created\t'Display created time'
 "
 complete -c exa -s 'm' -l 'modified'      -d "Use the modified timestamp field"
-complete -c -exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
+complete -c exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
 complete -c exa        -l 'changed'       -d "Use the changed timestamp field"
 complete -c exa -s 'u' -l 'accessed'      -d "Use the accessed timestamp field"
 complete -c exa -s 'U' -l 'created'       -d "Use the created timestamp field"
--- a/completions/completions.fish
+++ b/completions/completions.fish
@@ -65,6 +65,7 @@ complete -c exa -s 't' -l 'time'  -x -d "Which timestamp field to list" -a "
     created\t'Display created time'
 "
 complete -c exa -s 'm' -l 'modified'      -d "Use the modified timestamp field"
+complete -c -exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
 complete -c exa        -l 'changed'       -d "Use the changed timestamp field"
 complete -c exa -s 'u' -l 'accessed'      -d "Use the accessed timestamp field"
 complete -c exa -s 'U' -l 'created'       -d "Use the created timestamp field"
--- a/src/options/help.rs
+++ b/src/options/help.rs
@@ -39,28 +39,28 @@ FILTERING AND SORTING OPTIONS
                              date, time, old, and new all refer to modified.
 
 LONG VIEW OPTIONS
-  -b, --binary         list file sizes with binary prefixes
-  -B, --bytes          list file sizes in bytes, without any prefixes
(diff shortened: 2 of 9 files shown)


completions/completions.fish
```diff
@@ -65,6 +65,7 @@ complete -c exa -s 't' -l 'time'  -x -d "Which timestamp field to list" -a "
     created\t'Display created time'
 "
 complete -c exa -s 'm' -l 'modified'      -d "Use the modified timestamp field"
+complete -c -exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
```
ariasuni, line 68:
```suggestion
complete -c exa -s 'n' -l 'numeric-uid-gid' -d "List numeric user and group IDs."
```

src/options/help.rs
```diff
@@ -46,6 +46,7 @@ LONG VIEW OPTIONS
   -H, --links          list each file's number of hard links
   -i, --inode          list each file's inode number
   -m, --modified       use the modified timestamp field
+  -n, --numeric-uid-gid list numeric user and group IDs
```
ariasuni, line 49: There should two spaces between the option name and description:

```suggestion
  -n, --numeric-uid-gid  list numeric user and group IDs
```

This will be displayed with `exa --help`. Please align the descriptions to the same column.

src/options/view.rs
```diff
@@ -265,6 +266,14 @@ impl TimeFormat {
     }
 }
 
+impl UserFormat {
+    fn deduce(matches: &MatchedFlags<'_>) -> Result<Self, OptionsError> {
+        let flag = matches.has(&flags::NUM_UGID)?;
+        Ok(if flag { Self::Numeric } else { Self::Name })
+    }
+
+}
```
ariasuni, line 275: Just a style nitpicking:

```suggestion
}


impl UserFormat {
    fn deduce(matches: &MatchedFlags<'_>) -> Result<Self, OptionsError> {
        let flag = matches.has(&flags::NUM_UGID)?;
        Ok(if flag { Self::Numeric } else { Self::Name })
    }
}
```

ariasuni (changes requested):
The current behavior is correct and `-n` shouldn’t imply `-l`: in exa, options relative to a specific mode (long, grid, etc.) are ignored in different modes to allow overriding aliases.

I didn’t test your change, but the code looks good!

b05902132:
Thanks for your review! I applied all your suggested chanes. It's the first time I use this feature; not sure if I did it correctly though.

src/options/help.rs
b05902132: `--numeric-uid-gid` is too long to make the description align with other descriptions. Adding more spaces makes the alignment off even more. I will add a space to other options to make their descriptions align the description of `-n`.

b05902132:
Hi. I ran `exa --help` and found that the suggested change actually made the alignment worse. I added a new commit to fix that. 
Also I'd like to squash all suggested changes into one commit, but I'm new to open source world and not sure if it's allowed. I heard that you should not change published commits, but I don't think any one will base their work on my branch.

ariasuni (changes requested):
You should also add the option to the list of options that are useless without long for strict mode:
https://github.com/ogham/exa/blob/435161e6e4436f10e4faaa03c5f3deef26e12b36/src/options/view.rs#L87-L92

---

For tests, you should add the options in:
https://github.com/ogham/exa/blob/435161e6e4436f10e4faaa03c5f3deef26e12b36/src/options/view.rs#L343-L348

You can then copy
https://github.com/ogham/exa/blob/435161e6e4436f10e4faaa03c5f3deef26e12b36/src/options/view.rs#L553

and
https://github.com/ogham/exa/blob/435161e6e4436f10e4faaa03c5f3deef26e12b36/src/options/view.rs#L564
but use `-n` or `--numeric-uid-gid` instead.

ogham:
Hi, thanks for this! I can see the use for this option.

My feeling is that I'd rather call it just `--numeric` rather than `--numeric-uid-gid`. I can't think of any other columns where we display a name based on a number other than the user and group, and users will expect options with long names and hyphens in to be more complex, when this one is quite simple. There's also precedent: netstat uses `-n`, and GNU netstat uses `--numeric`.

Don't worry about the help text; keeping things in alignment is tricky when multiple people are making edits to the same file. It happens in code too. I'll sort it out when everything settles.

ogham:
> (I don't understand that section of code.)

Yeah, it was written when I was feeling more, uh... artistic. When I looked back on it recently I wished I'd just kept it simple. So don't worry if it was hard to follow.

> Also I'd like to squash all suggested changes into one commit, but I'm new to open source world and not sure if it's allowed. I heard that you should not change published commits, but I don't think any one will base their work on my branch.

There's no rule that applies to all projects, but _in general_, your second point is correct: if multiple people are using a branch then you should leave it alone, but if it's your own development branch that has not been merged yet, then feel free to squash.

b05902132:
> Hi, thanks for this! I can see the use for this option.
> 
> My feeling is that I'd rather call it just `--numeric` rather than `--numeric-uid-gid`. I can't think of any other columns where we display a name based on a number other than the user and group, and users will expect options with long names and hyphens in to be more complex, when this one is quite simple. There's also precedent: netstat uses `-n`, and GNU netstat uses `--numeric`.
It was following GNU `ls`. Since `exa` does not need to follow `ls`, I've updated the option to be `--numeric`.
> 
> Don't worry about the help text; keeping things in alignment is tricky when multiple people are making edits to the same file. It happens in code too. I'll sort it out when everything settles.
Okay.

PR #818, opened 2021-03-24
Update determine_time_zone according to GNU manual

### LINKED ISSUE #817: TZ environment variable that begins with colon is not properly read
`exa --version`
> exa v0.9.0

`echo $TZ`
> :America/New_York

`exa -l`
> Unable to determine time zone: No such file or directory (os error 2)

See here for details

https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html
> The third format looks like this:
> :characters
> If characters begins with a slash, it is an absolute file name; otherwise the library looks for the file /usr/share/zoneinfo/characters.

https://wiki.archlinux.org/index.php/Environment_variables#Examples
> TZ can be used to to set a time zone different to the system zone for a user. The zones listed in /usr/share/zoneinfo/ can be used as reference, for example TZ=":/usr/share/zoneinfo/Pacific/Fiji". When pointing the TZ variable to a zoneinfo file, it should start with a colon per the GNU manual.

PatriotRossii:
Fixes #817 

GNU manual: https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

> The third format looks like this:

> :characters
If characters begins with a slash, it is an absolute file name; otherwise the library looks for the file /usr/share/zoneinfo/characters.

### DIFF
--- a/src/output/table.rs
+++ b/src/output/table.rs
@@ -292,9 +292,21 @@ impl Environment {
 
 fn determine_time_zone() -> TZResult<TimeZone> {
     if let Ok(file) = env::var("TZ") {
-        TimeZone::from_file(format!("/usr/share/zoneinfo/{}", file))
-    }
-    else {
+        TimeZone::from_file({
+            if file.starts_with("/") {
+                file
+            } else {
+                format!("/usr/share/zoneinfo/{}", {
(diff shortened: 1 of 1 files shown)


ogham:
Excellent, thanks for this!

Issue #816, opened 2021-03-21
File permission in octal


TheFern2:
I might be the weird one out of the bunch, but I prefer to see file permissions in octal at times. For example if a tutorial is telling me to make sure a file has certain permissions 99% of the times is in octal. I did a very simple implementation of ls in python a while back that does this. Don't know if is a feature that people might like. If is something that you're not interested, please feel free to close the issue.

https://github.com/TheFern2/lspy

lx is an alias for ls.py which is installed in $HOME/.local/bin
```bash
~/g/fab-mentorship  (master) ❯❯❯ lx -ln
775 2 fernandob fernandob 4096 Jul 28 15:48 Career-Tips
775 3 fernandob fernandob 4096 Jul 28 15:48 Docs
775 3 fernandob fernandob 4096 Jul 28 15:48 Getting-Started
775 2 fernandob fernandob 4096 Jul 28 15:48 Images
775 4 fernandob fernandob 4096 Jul 28 15:48 Language-Specific
664 1 fernandob fernandob 1067 Jul 28 15:48 LICENSE
775 3 fernandob fernandob 4096 Jul 28 15:48 Misc
775 7 fernandob fernandob 4096 Jul 28 15:48 Projects
664 1 fernandob fernandob  246 Jul 28 15:48 README.md
775 3 fernandob fernandob 4096 Jul 28 15:48 Resources
775 2 fernandob fernandob 4096 Jul 28 15:48 Stories
775 2 fernandob fernandob 4096 Jul 28 15:48 Tech-Paths
775 5 fernandob fernandob 4096 Jul 28 15:48 Tools
~/g/fab-mentorship  (master) ❯❯❯ ls -l
total 52
drwxrwxr-x 2 fernandob fernandob 4096 Jul 28  2020 Career-Tips/
drwxrwxr-x 3 fernandob fernandob 4096 Jul 28  2020 Docs/
drwxrwxr-x 3 fernandob fernandob 4096 Jul 28  2020 Getting-Started/
drwxrwxr-x 2 fernandob fernandob 4096 Jul 28  2020 Images/
drwxrwxr-x 4 fernandob fernandob 4096 Jul 28  2020 Language-Specific/
-rw-rw-r-- 1 fernandob fernandob 1067 Jul 28  2020 LICENSE
drwxrwxr-x 3 fernandob fernandob 4096 Jul 28  2020 Misc/
drwxrwxr-x 7 fernandob fernandob 4096 Jul 28  2020 Projects/
[...]

PR #819, opened 2021-03-24
Add "octal-permissions" flag to list of options

PatriotRossii:
The absence of the flag was discovered thanks to #816

According to the
https://github.com/ogham/exa/blob/f42957fab8ef1f750315bd957a16baa848177f19/src/options/flags.rs#L65
and
https://github.com/ogham/exa/blob/86de17b788c676c08814441aac1becf2d2e40bdf/src/options/help.rs#L56
it really exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ These options are available when running with `--long` (`-l`):
 - **--git**: list each file’s Git status, if tracked or ignored
 - **--time-style**: how to format timestamps
 - **--no-permissions**: suppress the permissions field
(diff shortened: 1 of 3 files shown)


ariasuni:
Do you think you can add this flag to fish and zsh completion files, too?

PatriotRossii:
> Do you think you can add this flag to fish and zsh completion files, too?

You are right. Didn't know about them.

ariasuni:
Perfect, thanks a lot!

Issue #32, opened 2015-02-23
exa is not available on Windows


sirxyzzy:
You mentioned you don't have a Windows machine, I do, I tried to compile exa (without git, for now) and it looks as if a couple of types are not in std. or maybe they are, but the error message is hard to understand...

C:\Users\andy.cargo\registry\src\github.com-1ecc6299db9ec823\users-0.2.3\src\lib.rs:117:27: 117:32 error: unresolved import `libc::uid_t`. There is no `uid_t` in `libc`
C:\Users\andy.cargo\registry\src\github.com-1ecc6299db9ec823\users-0.2.3\src\lib.rs:117 use libc::{c_char, c_int, uid_t, gid_t, time_t};
                                                                                                                   ^~~~~
C:\Users\andy.cargo\registry\src\github.com-1ecc6299db9ec823\users-0.2.3\src\lib.rs:117:34: 117:39 error: unresolved import `libc::gid_t`. There is no `gid_t` in `libc`
C:\Users\andy.cargo\registry\src\github.com-1ecc6299db9ec823\users-0.2.3\src\lib.rs:117 use libc::{c_char, c_int, uid_t, gid_t, time_t};

---

Moderator edit: Windows support is being worked on in https://github.com/ogham/exa/pull/820

ogham:
There are currently three problems preventing Windows support:

- The [users](https://github.com/ogham/rust-users) crate uses the `uid_t` and `gid_t` types from libc, which are guaranteed to exist in Unix environments, but will not exist under Windows, which does users and groups differently. This is the error you're getting. I'm not sure how to adapt the users crate for Windows, but there would have to be equivalent functions of "find user information from ID" and "find list of users in a group" functions.
- The [locale](https://github.com/ogham/rust-locale) crate looks for Unix-style month and day names in the Unix path. This would also have to use the Windows API.
- The formatting codes in [ansi-term](https://github.com/ogham/rust-ansi-term) are designed for ANSI-style terminals, rather than Windows terminals. (This won't stop it from compiling, but the output would make it unusable)

I'm not familiar enough with Windows to know if the API offers the functionality. I mean, it probably does, but I'm not sure how to find it.

andyphilpotts:
Thanks for the pointers, this is most helpful


[...]

PR #820, opened 2021-03-26
Initial support for Windows

skyline75489:
Supports #32


skyline75489:
This PR is based on previous work by @zkat (see https://github.com/ogham/exa/issues/32#issuecomment-623042107). All the tests is green. This is what it looks like now:


exa-windows

There are still things that need to be fixed, of course. For example `git` does not seem to work. If the author want this, I'd like to continue improving the support for Windows.

skyline75489:
Hi @ariasuni  can I get a little help with the CI build? I've test it myself. It works on both Windows & Linux.

ariasuni:
I suppose the compilation error is due to the fact the CI uses Rust 1.42.0. Are you able to locally build exa with this version of Rust? If that’s the problem, you can just write two different `size` functions with the cfg attributes, or 

---

I suppose that afterwards I should set:

```
os:
  - linux
  - windows
```

in `.travis.yml` on master branch?

skyline75489:
Hi @ariasuni Thanks for the feedback! I've fixed the build. 

I'm like a true newbie of rust. Please double check to make sure I'm not doing anything stupid 😄

skyline75489:
As for the CI build, it would be nice to have a Windows job. Also I'm honestly OK with not having at the moment, since the Windows support is still very initial. Passing the tests does not really mean the experience on Windows is good enough.

djdv:
Just a heads up, the `--git` parameter works for conventional paths 🎉, but doesn't seem to work when the path is a UNC path (such as a network share).
I'm not familiar enough with `exa` (or Rust) to know where the path from `reorient` is being used, but it may be such that the prefix `\\?\...` has to be preserved in these cases, but stripped when the path is something like `\\?\X:\...`.
But that's just a guess.

Here is a side-by-side:
exa git

For context, `pwd.exe` is GNU's pwd and `pwd` is Powershell's own builtin.
(I should also say that for me personally this isn't a big deal, but I noticed the patch and wanted to check that case. Figured this would be good to know)

skyline75489:
thanks @djdv for the heads up. I do understand that the code currently won’t work with UNC paths and I plan to add support for UNC paths in the future. Right now I just want to build the basics things on Windows and start from there.

ogham:
This is really cool! Although the `--long` output looks really, really different from what I expect exa to look like :)

On that note, I think it's worth changing the name of the left-most column. I only named it "Permissions" because I knew its Unix output was going to be ten or more characters wide. Now it's just one, it looks strange to me with that long header name with all the space there — none of the other columns change their width so drastically when the header is added. Of course, that affects things like the `--no-permissions` argument, and the help text, so it's not quite as simple as changing it in one place.

I'm willing to add UNC path support later because supporting Windows is already such a big jump.

ogham:
About the use of Rust 1.42: the MSRV is specified in the README and the Travis check is just there to check it doesn't go out-of-date, but I'm not tied to that particular version. If you need a feature from a later one then I'd be happy to bump it.

skyline75489:
@ogham thanks so much for this amazing software. I’m finally able to have a unified ls experience across different systems and I am glad to help the community, even though i’m a rust newbie 😃

sergeevabc:
Err… Would you mind attaching a final Windows binary here or somewhere nearby, please?

skyline75489:
@ogham I have some thoughts about the permission as well. I would love to make it the same as `Mode` in PowerShell (see [link](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.1#examples). I assume this needs win32 APIs.

The timezone detection logic is also a missing piece. I haven't done the investigation but I assume we'll also need Win32 API for that.

skyline75489:
@sergeevabc I think in the future we can put a binary in the release page once the Windows support is merged and the next release is available. At the moment the Windows support is still initial. If you really want to try it, feel free to checkout my branch and compile it.

skyline75489:
OK I managed to implement "permission" to mimic the "mode" in PowerShell. Here's how it looks:

image

PowerShell gci result for reference:

image

ogham:
> on Windows the hidden attribute is not determined by the name (having dot prefix or not)

I think this comes down to which type of hidden file a user will come into contact with the most.

For example, macOS has the same feature — a literal "hidden file" attribute — that means non-dotfiles can still be hidden in the file explorer. However, the vast, _vast_ majority of hidden files will still be dotfiles, such as `.git`, `~/.local`, `~/.config`, and all the stuff that clutters your home directories. The convention seems to be:

- The file explorer hides both "hidden files" and dotfiles.
- `ls` without `-a` shows "hidden files", but still hides dotfiles.
- `ls -a` shows both "hidden files" and dotfiles.

So the question is, how common is it to encounter "hidden files" in Windows? If it's not very common, exa should show them by default. But if there are several popular applications that set the "hidden file" flag and expect to have it respected, then exa should hide them, and have `-a` show both "hidden files" and dotfiles.

I've used Windows before, but because I end up installing a load of Unix utilities that still use the dotfile convention, I have no idea how common it is for files to set the "hidden file" flag.

> the time zone detection logic needs to be implemented on Windows

exa uses the datetime library for this, which I also maintain. Unfortunately it's pretty Unix-centric (it assumes `/etc/localtime` will contain a link to a list of time zone offset tables). It could also do with some improvement — I won't blame you if you find the code hard to read or modify.

The two functions exa needs from Windows are:

1. Some way to get the current time zone (as some structure).
2. Some way to convert a FILETIME (such as its modified time) into a zoned time, using the structure from function 1.

Do you know if Windows has these exact functions, or will exa have to do something slightly differently?

Systemcluster:
>So the question is, how common is it to encounter "hidden files" in Windows

There are several files with the hidden flag created by default in the user profile directory,  e.g. the `AppData` directory or the `NTUSER.dat` file. `C:/ProgramData` is hidden as well and there are a number of programs that create other hidden directories under `C:/`. In `C:/Windows` there are also a bunch of hidden files.

All in all it's probably less common than under linux, but definitely existing.

ogham:
If it's that uncommon then I think the best thing to do is to mimic the behaviour of `ls` on macOS, hiding dotfiles but displaying "hidden files". And if this has unexpected side-effects there's still room to change it before v1.0.

ogham:
@Systemcluster you think that's the wrong way to do it?

I was thinking about the behaviour of tools like Git, which are used on Windows even if they aren't originally native to that environment, so they still create dotfiles and dot-folders.

skyline75489:
As @Systemcluster pointed out, it's not *very* common if you're only navigating your own files. But if you're working with Windows folders, the exa result might be a bit noisy comparing to `gci` and `Windows Explorer`. That's why I think this is something needs to be fixed. However, I don't really want to do it in this PR since it needs to modify the code in `main.rs` & possible various other places. I'd like to keep this PR as minimal as possible. 

For the time zone, there's the API [GetTimeZoneInformation](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-gettimezoneinformation) which should gives us what we want. Again, this would require win32 API crates and I don't want to make this PR bloated.

So generally I think this PR is ready now. I'm happy to open issues to track future works we mentioned in this thread.

src/output/render/permissions.rs
```diff
@@ -76,6 +88,33 @@ impl f::Permissions {
...
+impl f::Attributes {
+    pub fn render<C: Colours+FiletypeColours>(&self, colours: &C) -> Vec<ANSIString<'static>> {
+        let bit = |bit, chr: &'static str, style: Style| {
+            if bit { style.paint(chr) }
+              else { colours.dash().paint("-") }
+        };
+
+        vec![
+            bit(self.archive,   "a", colours.normal()),
+            bit(self.readonly,  "r", colours.user_read()),
```
skyline75489, line 100: I didn't create other color for `readonly`, `hidden` & `system`. It looks fine now. And I don't think people would actually complain the naming of the color, since only a handful people would actually read the code 😄  If someone does fancy dedicated colors on Windows, we can add them in the future.

src/fs/file.rs
```diff
@@ -354,6 +379,11 @@ impl<'dir> File<'dir> {
         }
     }
 
+    #[cfg(windows)]
+    pub fn changed_time(&self) -> Option<SystemTime> {
+        return self.modified_time()
```
skyline75489, line 384: I don't quite understand the difference between `changed_time` & `modified_time`. Nor do I know if `changed_time` exists on Windows. So I just did this.

src/fs/file.rs
```diff
@@ -420,6 +465,21 @@ impl<'dir> File<'dir> {
         }
     }
 
+    #[cfg(windows)]
+    pub fn attributes(&self) -> f::Attributes {
+        let bits = self.metadata.file_attributes();
+        let has_bit = |bit| bits & bit == bit;
+
+        f::Attributes {
```
skyline75489, line 473: The constants are grabbed from MSDN doc [here](https://docs.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants).

src/output/render/permissions.rs
```diff
@@ -76,6 +88,33 @@ impl f::Permissions {
...
+            bit(self.hidden,    "h", colours.special_user_file()),
+            bit(self.system,    "s", colours.special_other()),
+        ]
+    }
+
+    pub fn render_type<C: Colours+FiletypeColours>(&self, colours: &C) -> ANSIString<'static> {
```
skyline75489, line 106: This is to mimic the behaviour of PowerShell.

KnapSac:
I'd like to help out with the Windows support, so some tracking issues would be nice.

floh96:
> As @Systemcluster pointed out, it's not _very_ common if you're only navigating your own files. But if you're working with Windows folders, the exa result might be a bit noisy comparing to `gci` and `Windows Explorer`. That's why I think this is something needs to be fixed. However, I don't really want to do it in this PR since it needs to modify the code in `main.rs` & possible various other places. I'd like to keep this PR as minimal as possible.
> 
> For the time zone, there's the API [GetTimeZoneInformation](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-gettimezoneinformation) which should gives us what we want. Again, this would require win32 API crates and I don't want to make this PR bloated.
> 
> So generally I think this PR is ready now. I'm happy to open issues to track future works we mentioned in this thread.

@skyline75489 there is also a rust library for the win32 api now. See https://github.com/microsoft/windows-rs maybe it's possible to use it?

skyline75489:
@floh96 Yes that is one option to do this. I'm aware of the library provided by Microsoft. And I think it's a good choice. However calling the API isn't really enough. Will need to investigate more to find out how to make it work with exa code.

src/output/render/permissions.rs
ogham: Yeah, I think re-using existing colours for this is fine!

src/fs/file.rs
ogham: Ah, it's the time the file attributes change rather than the file's data. I don't think it does exist on Windows.

rivy:
Just FYI, files with leading `_` are also "hidden" by lots of older Windows applications. In the past, some built-in windows applications (eg, "File Explorer") refused to create files without a file "base" name seeing ".FILENAME" as a nameless file with only an extension. "_FILENAME" was used by various more POSIXy application ports as an alternative.

Prior art ([`msls`](https://github.com/u-tools/msls)) straddles the lines, hiding both ".FILENAME" and "_FILENAME" files as well as any files with the Windows hidden or system attributes. ".FILENAME" and "_FILENAME" are both somewhat odd filenames, so I think hiding them (on top of attribute HIDDEN and SYSTEM files) would be a good default for all systems. It hides files which IMO most users would expect to be hidden. Then, no extra flags or configuration would be needed. The usual `-a` could be used to display those hidden files.

rivy:
> So the question is, how common is it to encounter "hidden files" in Windows? If it's not very common, exa should show them by default. But if there are several popular applications that set the "hidden file" flag and expect to have it respected, then exa should hide them, and have `-a` show both "hidden files" and dotfiles.
> 
> I've used Windows before, but because I end up installing a load of Unix utilities that still use the dotfile convention, I have no idea how common it is for files to set the "hidden file" flag.

`exa` will likely be most used by developers who also install Windows and POSIXy utilities. Those (we?) folks will much more commonly encounter hidden files (and are also most likely to be annoyed by files they don't want to see at baseline).

rivy:
> > I think in the future we can put a binary in the release page once the Windows support is merged and the next release is available.
> 
> Yeah; I had to set up a Windows VM for testing one of my other projects. I can use it to compile Windows exa and make it part of the release when the time comes.

There's no need to set up a VM.

GitHub Actions can compile and upload binaries for releases automatically. I've helped build a few examples (see the [[`pastel`] CI](https://github.com/sharkdp/pastel/blob/master/.github/workflows/CICD.yml#L224-L232)). Once you're testing the code on GitHub Actions windows platforms, the binary upload for releases is straight forward and can be completely automatic.

skyline75489:
@rivy thanks for the heads up. Does the _ prefix only apply to file, or does is also apply to folders? 

I’m aware of the existence of _ prefix file back in the days dot prefix is not supported. But right now Windows is way more Unix-friendly and dot prefix works just fine. So I’m not really sure whether it’s relevant any more.

Systemcluster:
> @Systemcluster you think that's the wrong way to do it?

I think ultimately supporting the Windows-native hidden attribute would be keep the behavior consistent with the Windows Explorer and the Windows-native command line tools (like `dir` in CMD and `dir`/`Get-ChildItem` in PowerShell). Hiding dotfiles by default feels very foreign on Windows.

> I was thinking about the behaviour of tools like Git, which are used on Windows even if they aren't originally native to that environment, so they still create dotfiles and dot-folders.

I forgot to mention this in my previous list, but the `git init`/`git clone` created `.git` directory also properly has the hidden attribute set.

skyline75489:
@Systemcluster you made a good point about hiding dotfiles on Windows does not follow the behavior of tools like PowerShell. I found this immediately after playing around the -a flag. 

But I’d argue hiding dot files is actually what I *want* most of the time when using a terminal. Let’s face it, Windows has been and will probably be more Unix-friendly in the future. I don’t think there’s too much harm since the root of dotfiles *is* from Unix world. Dot prefix is not even supported in older versions of Windows, as mentioned in other discussions. So i wouldn’t feel hiding dotfiles makes exa “foreign”. But rather I think hiding dotfiles makes exa modern.

zkat:
I think hiding dotfiles AND Windows hidden files is the right way to go about it. Realistically, dotfiles are intended to be hidden, on any platform. That's their intention. So let's go with that.

rivy:
> @rivy thanks for the heads up. Does the _ prefix only apply to file, or does is also apply to folders?

Both files and directories.
 
> I’m aware of the existence of _ prefix file back in the days dot prefix is not supported. But right now Windows is way more Unix-friendly and dot prefix works just fine. So I’m not really sure whether it’s relevant any more.

There are still several ports that use them. Here are some examples from my Windows home directory...

```text
...
lrw------a  1 Roy                 0 Dec 13  2018 _curlrc -> .sh\.curlrc
-rw------a  1 Roy               924 Mar 16 09:19 _lesshst
lrw------a  1 Roy                 0 Dec 13  2018 _netrc -> .secrets\.netrc
lrwx------  1 Roy                 0 Dec 13  2018 _ssh -> .secrets\SSH\
-rw------a  1 Roy               16k May  5  2019 _viminfo
...
```

And they are just as annoying as dotfiles/directories when you see them cluttering up an otherwise pristine directory listing. 😄 

I agree with @zkat. To be specific, IMO, hiding dotfiles/directories, underscore-files/directories, AND all Windows hidden and system attributes files/directories is the best, "modern" way to go. Again, that way there is no need for additional configuration flags; only the `-a` is needed to show all hidden files. Otherwise, you'd need additional configuration flags to hide/show some subset of those files.

ogham:
Looks like there's going to be a bugfix `v0.10.1` release coming next. I'll continue looking at this PR once that release is made.

skyline75489:
I don't quite get it why CI started to fail after merging master... btw why do we still want to support 1.42 now that 1.49 is more "mainstream"

llacb47:
@skyline75489 I just `cargo install`ed your branch and the output looks wrong.

image

skyline75489:
@llacb47 Hi I'm mainly targeting to use `exa` in Windows Terminal. Yeah it does not seem to work in conhost (the traditional console used by `cmd.exe` & `powershell.exe`).

Also you'll need at least Windows 10 1809 for it to work correctly. Even though it should compiles & run on older version of Windows, it won't produce correct output because the lacking of ConPTY interface.

llacb47:
I'm on Windows 10 20H2. Any known reason why it doesn't work in conhost? `https://github.com/zkat/exa` does.

skyline75489:
Huh, that's even weirder. My branch is based on zkat's branch. 

I haven't got the time to investigate it yet. Thanks for all the information, though.

skyline75489:
@llacb47 I've fixed this. It now works with traditional console. Give it a try if u got a minute 😄

llacb47:
It works! Thanks a lot @skyline75489

ariasuni:
Note that the CI is failing because Rust depends on git2 and that its latest version needs Rust 1.45.0. I just merged #854 that fixes this problem, so you can rebase on master (it’s good to check that with your work, exa still compiles on Linux).

Hopefully, the CI won’t annoy again ^^’

skyline75489:
Thanks @ariasuni for the information. The CI is green again. 

I do try to make the PR harmless to other systems, and use local build of exa on my Linux machines to ensure that.

skyline75489:
OK another unfortunate issue from old compiler (1.45.2):

```text
bits.push(self.colours.symlink_path().paint(String::from(std::path::MAIN_SEPARATOR)));
    |                                                         ^^^^^^^^^^^^ the trait `std::convert::From<char>` is not implemented for `std::string::String`
````

All I want it a separator based on the OS running...

**Comment by ariasuni** [comment]
@skyline75489 You can probably use `.to_string()` though. I’ll think about bumping a slightly more recent version so that it doesn’t annoy you again.

**Comment by skyline75489** [comment]
@ariasuni Thanks, as always. The `to_string` way works.

**Comment by thedeveloper-sharath** [comment]
@ogham @skyline75489  I have been testing this branch on my windows machine and everything seems to be working fine. Atleast for my use case except the grid view only works in some places, is there a limit on the number of files for grid view to be activated.

I am using the below function in powershell to call this

image

image

**Comment by ark-** [comment]
For anyone who wants to try this, I used the following one-liner.
```
cargo install --no-default-features --git https://github.com/skyline75489/exa.git --branch chesterliu/dev/win-support
```

I also saw the same issue described above by @thedeveloper-sharath 

> 
> 
> @ogham @skyline75489 I have been testing this branch on my windows machine and everything seems to be working fine. Atleast for my use case except the grid view only works in some places, is there a limit on the number of files for grid view to be activated.
> 
> I am using the below function in powershell to call this
> 
> image
> 
> image

No other issues noticed.

**Comment by rashil2000** [comment]
Just a small heads up, exa throws up errors when compiled from this branch:

image

This only happens in root of drives (C:\\, D:\\), for all other directories it works fine.

**Comment by skyline75489** [comment]
@rashil2000 yeah that is expected. This is not even exa's issue. The filesystem API Rust uses can't return the needed information for admin's directory when you're an unprivileged user. I don't quite know how to mitigate this, though.

**Comment by rashil2000** [comment]
Some errors shows up even when you're running as admin:

image

One simple workaround can be to just disable the checks (which result in this error) when exa detects that it is in root of drives (C:\\, D:\\). Only showing the name in those cases should suffice.

**Comment by zkat** [comment]
So uhhhh.. is there any chance we can merge this now and fix the corner cases later? :P

**Comment by daviessm** [comment]
@skyline75489 Is there any chance you could test that my PR https://github.com/ogham/exa/pull/941 doesn't break on Windows? I don't have a Windows environment set up at the moment but could probably figure one out if my changes break things.

**Comment by skyline75489** [comment]
@daviessm there's no merge conflict and only a small changeset is needed to make it work on Windows:

```diff
diff --git a/src/fs/file.rs b/src/fs/file.rs
index 959ff9b..30c9cd8 100644
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -238,6 +238,7 @@ impl<'dir> File<'dir> {
     }
 
     /// Whether this file (directory) is a `btrfs` subvolume
+    #[cfg(unix)]
     pub fn is_btrfs_subvolume(&self) -> bool {
         if cfg!(unix) {
             if self.is_directory() &&
diff --git a/src/main.rs b/src/main.rs
index cbc08eb..5627390 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,6 +33,7 @@ use std::io::{self, Write, ErrorKind};
 use std::path::{Component, PathBuf};
 
 use ansi_term::{ANSIStrings, Style};
+#[cfg(unix)]
 use proc_mounts::MOUNTS;
 
 use log::*;
@@ -54,29 +55,28 @@ mod theme;
 lazy_static! {
     static ref ALL_MOUNTS: HashMap<PathBuf, MountedFs> = {
         let mut m = HashMap::new();
-        if cfg!(unix) {
-            for mount_point in &MOUNTS.read().unwrap().0 {
-                let mount = MountedFs {
-                    dest: mount_point.dest.to_string_lossy().into_owned(),
-                    fstype: mount_point.fstype.clone(),
-                    source: mount_point.source.to_string_lossy().into_owned(),
-                    subvolume: match mount_point.fstype.as_str() {
-                        "btrfs" => {
-                            let mut subvol = None;
-                            for option in &mount_point.options {
-                                if option.starts_with("subvol=") {
-                                    subvol = Some(option[7..].to_string());
-                                    break;
-                                }
+        #[cfg(unix)]
+        for mount_point in &MOUNTS.read().unwrap().0 {
+            let mount = MountedFs {
+                dest: mount_point.dest.to_string_lossy().into_owned(),
+                fstype: mount_point.fstype.clone(),
+                source: mount_point.source.to_string_lossy().into_owned(),
+                subvolume: match mount_point.fstype.as_str() {
+                    "btrfs" => {
+                        let mut subvol = None;
+                        for option in &mount_point.options {
+                            if option.starts_with("subvol=") {
+                                subvol = Some(option[7..].to_string());
+                                break;
                             }
-                            subvol
-                        },
-                        _ => None,
-                    }
-                };
-                m.insert(mount_point.dest.clone(), mount);
-            }
-        }
+                        }
+                        subvol
+                    },
+                    _ => None,
+                }
+            };
+            m.insert(mount_point.dest.clone(), mount);
+        };
         m
     };
 }
diff --git a/src/output/file_name.rs b/src/output/file_name.rs
index cd9e50a..8849497 100644
--- a/src/output/file_name.rs
+++ b/src/output/file_name.rs
@@ -358,6 +358,7 @@ impl<'a, 'dir, C: Colours> FileName<'a, 'dir, C> {
 
         match self.file {
             f if f.is_mount_point()      => self.colours.mount_point(),
+            #[cfg(unix)]
             f if f.is_btrfs_subvolume()  => self.colours.btrfs_subvolume(),
             f if f.is_directory()        => self.colours.directory(),
             #[cfg(unix)]

```

daviessm:
@skyline75489 Thanks very much. I force-pushed the branch to https://github.com/daviessm/exa/commit/823764aaa9e6098a35b4d98ce6547fc4e2a0eeba if you'd like to re-check.

src/output/file_name.rs
```diff
@@ -301,11 +315,16 @@ impl<'a, 'dir, C: Colours> FileName<'a, 'dir, C> {
 
         match self.file {
             f if f.is_directory()        => self.colours.directory(),
+            #[cfg(unix)]
```
BlackHoleFox, line 318: Windows has executable files too?

src/output/table.rs
```diff
@@ -116,12 +124,18 @@ pub enum Column {
     Permissions,
     FileSize,
     Timestamp(TimeType),
+    #[cfg(unix)]
     Blocks,
+    #[cfg(unix)]
     User,
+    #[cfg(unix)]
     Group,
+    #[cfg(unix)]
```
BlackHoleFox, line 133: Windows [has hard links](https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions) too

BlackHoleFox:
Just a passerby who's interested in seeing this land :) I built this branch and it seems to be working great for me. Symlinks and junctions are recognized correctly but hardlinks seem to not appear with the `--links` option.

s5bug:
What's the best way to build this / best commit to build with right now? Will it work with WinBTRFS?

daviessm:
@s5bug I just checked out the branch this pull request is made from and it works fine with WinBTRFS on my installation. There's no reason it should care about the filesystem as it uses OS calls to traverse the fs.

Note though that my PR #941 for highlighting BTRFS subvolumes won't work on Windows if/when merged because it uses a *nix-y way of finding mount points. I can try to add Windows support if you're interested.

s5bug:
I don't actually use any of the fancy features in WinBTRFS, so it isn't a big deal for me 😄 Thank you!

floh96:
@ogham friendly ping could this pr get a review?

ariasuni:
Well, I don’t think we should wait on @ogham. I’ll see if I can review it sometime before the end of this month… but if the basic features work for people on Windows with this build, it should be fine to merge.

Aankhen:
> Well, I don’t think we should wait on @ogham. I’ll see if I can review it sometime before the end of this month… but if the basic features work for people on Windows with this build, it should be fine to merge.

In case it helps, I’ve been using the branch on Windows 10 for months and it seems to work well. I stopped using the builtin entirely.

rashil2000:
> In case it helps, I’ve been using the branch on Windows 10 for months and it seems to work well. I’ve stopped using the builtin dir entirely (both PowerShell’s and Command Prompt’s).

I have been using it for this long too. Certainly replaces Cmd's `dir` builtin, but it can't really replace PowerShell because the latter returns objects/pipelines ;)

(if it's helpful, PowerShell's builtin `dir` can have icons too, using the [Terminal-Icons](https://github.com/devblackops/Terminal-Icons) module, which is also cross-platform)

Aankhen:
Sorry, to clarify, I use exa instead of <kbd>dir</kbd> when I want to see a file listing. I wouldn’t expect to get rich objects from it in PowerShell, heh.

skyline75489:
As the author, just want to add that I've also been using my personal build of `exa` on several Windows machines since the birth of this PR and it suits me quite well. 

I'll be here if any help is needed to get this merged :)

rashil2000:
> As the author, just want to add that I've also been using my personal build of `exa` on several Windows machines since the birth of this PR and it suits me quite well.
> 
> I'll be here if any help is needed to get this merged :)

Would it be possible to address the comments regarding executables and hard links?

skyline75489:
@rashil2000 yeah I see those comments, but it's been a very long since I initially wrote this PR and right now I don't really want to mess with the code again to add *more* features in it. It's already a big & messy PR that's somewhat difficult to review. We can always file more issues in the future, can't we.

src/fs/file.rs
ariasuni: Could this piece of information and MSDN link be added to the source code, as a comment?

src/output/file_name.rs
ariasuni: It doesn’t seem like there’s a flag for executable files on Windows; my understanding is that contrarily to Linux where we need to set a file as executable, under Windows any normal file could be potentially executed, and the `.exe` (or `.msi` etc.) file extension is the only thing that can tell it’s supposed to be executed.

src/output/render/permissions.rs
ariasuni: Not sure what you mean by that? Not sure what this code does, but it’s worthy of a comment on GitHub, maybe it’s worthy to explain the thing quickly in a comment (at least it’s not obvious what’s going on for a non-Powershell user, I’d say).

rashil2000:
Any updates on this? What's the blocker?

ariasuni:
@rashil2000 Well, this is a big PR, and that I really know anything about Windows. Also, I’d like to finish some work on the CI so that we can then add checks that the Windows version compiles, so that it doesn’t break after being merged.

I know this is frustrating but I have a very limited bandwidth right now. I triaged and answered to all new issues and comments that needed it or that I could answer right away, and merged some PRs, because I took vacations and use that time to *work* on exa instead of doing something else.

What this project need is more maintainers (that is, more than one), and reviews from other people would help immensely as well. Anyway, I’ll try to merge this PR soon, but you know how it goes, sometimes there are other priorities.

garyo:
Can anyone just post a binary for Windows somewhere in the meantime?

src/output/render/permissions.rs
skyline75489: Oh i mean this is to mimic the behaviour of `gci` command in PowerShell, which is the builtin `ls` alternative in PowerShell.

Freed-Wu:
Screenshot from 2022-05-14 21-36-46

Screenshot from 2022-05-14 21-47-22

It looks exa cannot display group/user in windows? And the time is for UTC not localtime. The font is different. Seem lsd can display correct previledge and directory size in 2rd picture.

However, display h/s (hidden/system) is interesting. I hope exa can enter the msys2/mingw, although I failed to makepkg-mingw it...

Freed-Wu:
See bug <https://github.com/Peltoche/lsd/issues/658>, exa is better than lsd in performance.

Freed-Wu:
And `exa --color auto` always behave `never` in windows. `lsd` is same. I am not sure if it is a bug. See <https://github.com/Peltoche/lsd/issues/659>

Freed-Wu:
If a symlink is invalid, exa will corrupt in windows.

```
ln -s non_exist symlink
exa symlink
```

get `os error 267`

ShaneMurphy2:
Also been using this branch, seems pretty stable for most use cases. Though I second that the time is in UTC.

src/output/file_name.rs
skyline75489: Yea I agree with @ariasuni on this one. Detecting executables seems to be purely relying on file extensions, which is not elegant (orreliable)

tigercat2000:
Hey, I have some experience with the windows API from working on a very similar project for fun;
So I went ahead and added support for groups and users using the `windows` crate here:
https://github.com/tigercat2000/exa/tree/win-support

Screenshot showing exa displaying groups and users in C:\\Windows\\System32

I'm not sure what the standard practice around here is for contributing to open pull requests, so I figured I would just leave a link to my branch here and see what y'all think :)

BlackHoleFox:
> So I went ahead and added support for groups and users using the windows crate here:

imo this should probably be off by default, behind a CLI flag. Unlike on UNIX, obtaining the security group info on Windows is incredibly expensive/slow with not much room for optimization. The same thing happens with `lsd`.

tigercat2000:
> > So I went ahead and added support for groups and users using the windows crate here:
> 
> imo this should probably be off by default, behind a CLI flag. Unlike on UNIX, obtaining the security group info on Windows is incredibly expensive/slow with not much room for optimization. The same thing happens with `lsd`.

It currently only fetches this data if the renderer requests it from File::user() or File::group(), implying the user is already asking for user/group information- is there a point to going beyond that? Perhaps turning off the User column by default in `-l` mode?

AceofSpades5757:
@tigercat2000 you would want to create a pull request.

Is there something missing for Windows support to be added? I looked through a good bit of the comments and it looks like these are still the pending issues.

* executables can only be detected using the file extension
* hard links
* symlink will corrupt on Windows
* Time is in UTC

tigercat2000:
@AceofSpades5757 
> @tigercat2000 you would want to create a pull request.

My branch is a modification of this PR's branch, thus it either repeats the commits from here if I do it against `exa:master`, or relies on this PR's author if I do it against `skyline75489:chesterliu/dev/win-support`. Or, I could wait until this PR is merged at some point and contribute just my new work to `exa:master`.
Neither one seems ideal, honestly. I'd prefer to hear from @ogham for the first, or @skyline75489 for the second.

Gideon-Felt:
What's the progress/status on resolving all issues prior to merge?

AceofSpades5757:
Do you know why `~` doesn't work on Windows? It works on Linux and using other commands like `dir` or `Get-Item`.

rashil2000:
> ... and using other commands like `dir` or `Get-Item`.

This works only in PowerShell because it has special case handling for ~ character. 

Otherwise ~ has no meaning on Windows.

scottbilas:
> Do you know why `~` doesn't work on Windows? It works on Linux and using other commands like `dir` or `Get-Item`.

You can emulate unixy shell expansion with a helper. Here's some copy pasta from my profile, maybe it will be helpful to you.

```posh
function Expand-Path {
    [CmdletBinding()]
    param(
        [Parameter(Mandatory=$True, ValuefromPipeline=$True)]
        [string[]]$path
    )

    process {
        foreach ($i in $path) {
            if ($i.StartsWith('~')) {
                $Home + $i.Substring(1)
            }
            else {
                $i
            }
        }
    }
}

function l { exa --all --group-directories-first --icons --classify ($args | Expand-Path) }
function ll { l --long --git --header ($args | Expand-Path) }
```

scottbilas:
@skyline75489 - first of all, thanks. :D I'm already running happily with your branch.

Second, a small bug report. `exa /` gives me some errors:

```
$ exa /
[/Config.Msi: Access is denied. (os error 5)]
[/DumpStack.log: Access is denied. (os error 5)]
[/DumpStack.log.tmp: The process cannot access the file because it is being used by another process. (os error 32)]
[/hiberfil.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[/pagefile.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[/swapfile.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[/System Volume Information: Access is denied. (os error 5)]
$RECYCLE.BIN
$WinREAgent
boot
bootTel.dat
builds
dks
Documents and Settings
Intel
OneDriveTemp
PerfLogs
ProgramData
...
```

Qyriad:
> Otherwise ~ has no meaning on Windows.

It might be worth noting that `~` has no special meaning on Linux either. Posix-y shells (bash, zsh, etc) and file dialogs expand `~` to your home directory by convention (and Powershell also follows this convention) but it doesn't have special meaning to Linux itself, just like with Windows.

tigercat2000:
@scottbilas Interesting note, `ls` under WSL2 can handle the directories, but not the files.

scottbilas:
> ...and Powershell also follows this convention...

Powershell has some weak support for a couple shell-expansion-ish features (like ~ and *), but a crucial limitation is that it only does this on internal commands, and not on processes it launches. I'd guess this is because of compatibility with Windows native cli tools that are expecting to get raw wildcards in their args.

Anyway, the bottom line is that this kind of thing does not work on Windows:

exa ~
exa /Prog*
exa FooB?r

(Let's not even mention the rich expressions that a shell like zsh supports..)

Any port of a unix-first CLI tool to Windows will hit this impedance mismatch and then be stuck with either a) duplicating a portion of bash's basic expansion features like globbing, just for Windows (yuck), or b) leave the Windows port with an unhappy surprise ("huh, I can't do `exa *.cpp`??")

I guess the "right" solution would be to find a Powershell module that can glob-expand, and direct users to write little helper functions that route calls through that to unix-first CLI tools where such processing makes sense (like exa).

scottbilas:
> @scottbilas Interesting note, `ls` under WSL2 can handle the directories, but not the files

Yeah same here. Never noticed it before..

A tiny bit more info on the problem at https://github.com/athityakumar/colorls/issues/297#issuecomment-811970546

garyo:
Using a shell-glob library (e.g. https://github.com/p-ranav/glob) is a fine idea, but note that some users (like me) will almost always be using it from within msys or cygwin anyway, and running a "real" shell, so in that case the current version is fine because the shell expands the wildcards.

amrbashir:
there is also the [`wild`](https://crates.io/crates/wild/2.1.0) crate which mimics the unix expansion for windows.

src/fs/dir.rs
```diff
@@ -111,6 +111,13 @@ impl<'dir, 'ig> Files<'dir, 'ig> {
                     continue;
                 }
 
+                // Also hide _prefix files on Windows because it's used by old applications
+                // as an alternative to dot-prefix files.
+                #[cfg(windows)]
+                if ! self.dotfiles && filename.starts_with('_') {
```
garyo, line 117: I think this is overdoing it. Python uses `__init__.py` for top-level modules, and this hides all such files. If you're running this build with msys/cygwin, most likely your dotfiles actually start with '.' anyway.

AceofSpades5757: I don't know any files, except for `_vimrc` on Windows that follows this convention either.

ntpeters: > If you're running this build with msys/cygwin, most likely your dotfiles actually start with '.' anyway

Even without msys/cygwin. I don't mess with those and only use pwsh, and just use the normal dot prefix for all my dotfiles.

bitzl (approved):
Looks good to me. I think this should be merged as is, and if there might be any further comments or issues, these should be a new PR since this one is already huge.

ariasuni:
OK, I will merge it then. Please open issues for unfinished, buggy or missing stuff, so that we can improve Windows support!

AntoniosBarotsis:
@ariasuni About the failing workflow, [`cargo-hack`](https://github.com/taiki-e/cargo-hack) seems to specify `rustc v1.60`, you are using [`1.56`](https://github.com/ogham/exa/blob/f3ca1fe6f7d8242cee3d9fd602c96c9db041a7a0/.github/workflows/unit-tests.yml#L31).

ariasuni:
Yeah, there’s #1125 for the upgrade of the Rust version (if it takes too long or if I have the motivation to open my code editor soon, I’ll make my own commit). Using `fail-fast: false` wouldn’t change a thing since there’s only one job each time.

AntoniosBarotsis:
`fail-fast` should make all jobs run regardless of whether the first one failed. By all jobs I mean the same tasks in the different OSes you defined in your matrix.

eggbean:
I can't see this mentioned anywhere, but I apologise if it has: when piping to less or whatever, exa on Windows always shows colour, no matter what options are used.

specter119:
@eggbean `exa --color=never` got the no color result on my windows machine. however, I sitll build based on https://github.com/skyline75489/exa/tree/chesterliu/dev/win-support

eloc3147:
@eggbean I saw your comment in #1133
When you're piping into `less`, are you running in mingw? I don't have `less` in command prompt or powershell

ntpeters:
I also can't repro the color issue when piping to less on Windows.
I'm running the Windows build of less installed via scoop from [here](https://github.com/jftuga/less-Windows).

eggbean:
> @eggbean I saw your comment in #1133
> 
> When you're piping into `less`, are you running in mingw? I don't have `less` in command prompt or powershell

I'm using cmd.exe (with clink) and (scoop shim set to) the ls from the uutils-coreutils scoop package.

eggbean:
> @eggbean I saw your comment in #1133
> 
> When you're piping into `less`, are you running in mingw? I don't have `less` in command prompt or powershell

I'm using cmd.exe (with clink) and the `less` version in busybox (scoop package).

eloc3147:
I also cannot reproduce this color issue using:
- Clink v0.4.9 [git:2fd2c2]

- Less from BusyBox v1.37.0-FRP-5007-g82accfc19 (2023-05-28 11:00:58 BST)
(mingw64-gcc 12.2.1-8.fc38; mingw64-crt 10.0.0-4.fc38; glob)

- Exa c697d066702ab81ce0684fedb4c638e0fc0473e8

`exa` and `exa --color=never` both do not emit color codes to less, only `exa --color=always` does.

eggbean:
I probably have an older build?

```
C:\Windows\System32>exa --version
v0.10.1 [+git]
https://the.exa.website/
```

I got it from someone here, downloaded from their Google Drive. Could I try what you have there, or where did you get it from?

eggbean:
If I install rust, how do I get the relevant code to compile?

eloc3147:
@eggbean 
That was my google drive link ;) I've tested that (older) build and it also behaves correctly when piped.

To build:
- Clone this repo `git clone https://github.com/ogham/exa`
- Open the repo dir `cd exa`
- If you want to just build: `cargo build`
- If you want to build and install into your path: `cargo install --path .`

No magic here, just a standard rust project build. No special setup should be needed.

If you can repro this issue with your own build, it's probably time to open a dedicated issue.

eggbean:
Ha, yes it was you 🤭

I have a bit of a low-disk space situation going on with my Windows laptop and haven't taken the time to sort it out, hence the delay.

How to you even install cargo on Windows? I have found and installed rust with `winget install -e Rustlang.Rust.MSVC` but it doesn't give me cargo and I cannot find it in the winget repository. Would prefer to use winget if I can.

eloc3147:
The recommended way of installing rust is with [rustup](https://rustup.rs/)

I find installing programming languages with the OS supplied package manager is almost always a bad idea, but if you must use winget, the package you want is `Rustlang.Rustup`, this will install all the default tools and allow you to manager your rust version(s).

mattdkerr:
same on:
- v0.10.1 [+git] 
- Windows 11
- built from cargo command
```
[.\DumpStack.log.tmp: The process cannot access the file because it is being used by another process. (os error 32)]
[.\hiberfil.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[.\pagefile.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[.\Recovery: Access is denied. (os error 5)]
[.\swapfile.sys: The process cannot access the file because it is being used by another process. (os error 32)]
[.\System Volume Information: Access is denied. (os error 5)]
```

AndreM222:
how is the setup for windows powershell and will it work with terminal-icons?

ariasuni:
exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and Windows support is better in [the active fork eza](https://github.com/eza-community/eza), please check it out.

PR #822, opened 2021-03-27
Fix a rounding error in human readable filesizes

### LINKED ISSUE #812: Incorrect FIle size with `exa -l`
version: `exa v0.9.0`

how it can be reproduced:
create a 32G file on disk and run `exa -l` on this file

unexpected result: 
it says the file is with the size of 34 GB instead of 32 GB.

something must be wrong with the size calculation, I presume.

ariasuni:
Fix #812

For example, for a file of 454 908 bytes, exa displayed 454k but now displays 455k.

This was due to the fact that when the size of the file in the appropriate base (k, M, etc.) was superior to 10, exa stopped displaying the first decimal and formatted the number as an int by casting the float, which in practice was the same thing as if it took the floor of the size. Now, exa uses the same code path that does that the round case, except that it tells the library it uses to use 0 decimals in this particular case.

### DIFF
--- a/src/output/render/size.rs
+++ b/src/output/render/size.rs
@@ -41,9 +41,9 @@ impl f::Size {
         };
 
         let symbol = prefix.symbol();
-        let number = if n < 10_f64 { numerics.format_float(n, 1) }
-                              else { numerics.format_int(n as isize) };
-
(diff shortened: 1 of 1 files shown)


ogham:
Very well spotted, thank you.

PR #830, opened 2021-04-06
Use thousand separators again and fix alignement when it’s not ASCII

### LINKED ISSUE #657: Column alignment ruined by certain file sizes in binary mode
example: `exa -b -l`
```
.rw-rw-r--@      30Ki xyz 19 apr 23:02 msvcp140_1.dll
.rw-rw-r--@     200Ki xyz 19 apr 23:02 msvcp140_2.dll
.rw-rw-r--@ 1 002Ki xyz 19 apr 23:02 ucrtbase.dll
.rw-rw-r--@     357Ki xyz 19 apr 23:02 vccorlib140.dll
.rw-rw-r--@      83Ki xyz 19 apr 23:02 vcruntime140.dll
```
Sizes between 1000 and 1023 (with size suffix) have this problem. It seems that it compensates for the wider size, but in the "opposite direction".

Apologies for the windoze-related example. :)

ariasuni:
Fix #657

- If the number has no prefix (Ki, Mi, etc.), use `format_int()` instead of `to_string()` to use thousand separator
- Use `DisplayWidth::from()` on strings instead of assuming ASCII.

I introduced a regression in #822: I used `format_float()` instead of `format_int()` in certain cases, but the first doesn’t include thousand separators in its representation, so it fixed the alignment issue but now the numbers aren’t displayed properly.

So, the initial bug was that when using `-b`, sometimes the size of a file is for example between 1000KiB and 1024KiB, so it’s under 1MiB and it’s the only case where thousands separator were used (since no thousand separator is used with `-B`). Size of numbers was computed by using stdlib `.len()`, but the thousand separator is non ASCII for some languages (e.g., French or Swedish, which both uses U+202F). This problem was present except when the size was between 1000 and 1024 bytes because then it would use another code path with a plain `.to_string()`, which was obviously wrong as well.

Just for fun I found where the comment about how «it’s guaranteed to be ASCII» has been introduced: 4c2bf2f2e6231c14a8aa44cb49c828034c342a69. It seems we can’t assume anything we don’t write ourselves is ASCII after all.

Before (0.9.0, `LC_NUMERIC=sv_SE.UTF-8`):
```
$ exa -lb
.rw-r--r--     1 002 ariasuni  6 Apr 15:41 1000_bytes.txt
.rw-r--r-- 1 014Ki ariasuni  6 Apr 15:39 1000_Ki.txt
```

Before (0.10.0):
```
$ exa -lb
.rw-r--r--   1002 ariasuni  6 Apr 15:41 1000_bytes.txt
.rw-r--r-- 1014Ki ariasuni  6 Apr 15:39 1000_Ki.txt
```

After (`LC_NUMERIC=en_US.UTF-8`):
```
$ exa -lb
.rw-r--r--   1 002 ariasuni  6 Apr 15:41 1000_bytes.txt
.rw-r--r-- 1 014Ki ariasuni  6 Apr 15:39 1000_Ki.txt

```

After (`LC_NUMERIC=sv_SE.UTF-8`):
```
$ exa -lb
.rw-r--r--   1 002 ariasuni  6 Apr 15:41 1000_bytes.txt
.rw-r--r-- 1 014Ki ariasuni  6 Apr 15:39 1000_Ki.txt
```

---

Following this:
- Is it voluntary that sizes displayed when the `-B` flag don’t use a thousand separator?
- I’m wondering if `TextCell` shouldn’t have a method that takes a `contents` vec and compute the `DisplayWidth` by itself, instead of relying on every part of the code that needs something more complex than `TextCell::paint()` to do the right thing.

### DIFF
--- a/src/output/render/size.rs
+++ b/src/output/render/size.rs
@@ -36,20 +36,20 @@ impl f::Size {
         };
 
         let (prefix, n) = match result {
-            NumberPrefix::Standalone(b)   => return TextCell::paint(colours.size(None), b.to_string()),
+            NumberPrefix::Standalone(b)   => return TextCell::paint(colours.size(None), numerics.format_int(b)),
             NumberPrefix::Prefixed(p, n)  => (p, n),
         };
 
         let symbol = prefix.symbol();
-        let decimal_to_diplay = if n < 10_f64 { 1 } else { 0 };
-        let number = numerics.format_float(n, decimal_to_diplay);
+        let number = if n < 10_f64 {
+            numerics.format_float(n, 1)
+        } else {
+            numerics.format_int(n.round() as isize)
+        };
         
-        // The numbers and symbols are guaranteed to be written in ASCII, so
-        // we can skip the display width calculation.
-        let width = DisplayWidth::from(number.len() + symbol.len());
-
         TextCell {
-            width,
+            // symbol is guaranteed to be ASCII since unit prefixes are hardcoded.
+            width: DisplayWidth::from(&*number) + symbol.len(),
             contents: vec![
                 colours.size(Some(prefix)).paint(number),
                 colours.unit(Some(prefix)).paint(symbol),


ogham:
Thanks for this, as always, as well as the detailed explanation. I really appreciate it.

It seems like there are some test cases missing here. There are tests for `--bytes` and `--binary` flags, and there are tests for locales, but the locale tests don't check file sizes (they check dates), and I can't find any tests for sizes with both separators and symbols, which is why it could slip through and make it into a release.

> Is it voluntary that sizes displayed when the -B flag don’t use a thousand separator?

I don't think so. It certainly _feels_ like an oversight.

> I’m wondering if TextCell shouldn’t have a method that takes a contents vec and compute the DisplayWidth by itself, instead of relying on every part of the code that needs something more complex than TextCell::paint() to do the right thing.

From what I remember, the original intent was to allow for optimisations where we know the length in advance, without having to calculate it by UTF-8-parsing the string. However, I get that it's a _tiny_ optimisation. I doubt runtime would be affected at all with a method like that.

PR #836, opened 2021-04-09
Fix `EXA_GRID_ROWS` not working in some cases 

### LINKED ISSUE #835: `EXA_GRID_ROWS` does not work
The `EXA_GRID_ROWS` environment variable no longer works at all in v0.10.0. It worked in v0.9.0.

Bisection returned the commit 0f414cf0b9b75e948424ae153a50f6f2b489a63c.

To reproduce:
1. Create a directory with 2 files in it
2. `export EXA_GRID_ROWS=8`
3. `exa -Gl`
    → The files are displayed in a grid with 1 row

exa version: built from master (2aaead1)
OS: Pop!_OS 20.10

ariasuni:
Fix #835

In #790, I introduced an [early return](https://github.com/ogham/exa/pull/790/files#diff-4b2bc2ee82b980f63ad43373e8899a9527b74ab163a5317e2b00d975fd16061aR178) that short-circuits [the `EXA_GRID_ROWS` logic](https://github.com/ogham/exa/pull/790/files#diff-4b2bc2ee82b980f63ad43373e8899a9527b74ab163a5317e2b00d975fd16061aR186), so in some cases (namely, when the grid can fit in one row), `EXA_GRID_ROWS` is ignored.

This PR fixes that and introduces xtests to avoid future regressions. I also changed a comment to make it clear where’s the logic to handle `EXA_GRID_ROWS`.

I’m still not sure if the header line should be counted as an output row; right now, there’s no specific logic for when the option is used, so it’s counted as such.


ogham:
Excellent, thank you, both for the fix and for the detailed description. I totally thought there were xtests for `EXA_GRID_ROWS` already — I was surprised to find there wasn't.

> I’m still not sure if the header line should be counted as an output row; right now, there’s no specific logic for when the option is used, so it’s counted as such.

I think we should keep it as it is, counting the header line as a row, because the option is based on how much space gets taken up on the screen, and the header row takes up space like any other.

PR #837, opened 2021-04-09
Fix month name not following locale when date is in current year

### LINKED ISSUE #833: Time formating for the months does not follow the locale
### The version of exa being used
`v0.10.0`

### The command-line arguments you are using
```
LC_TIME=fr_FR.UTF-8 exa -l /
```

### The issue
exa-report

Since v0.10.0, in the long format, with the time-style set to default, exa only respects the locale for timestamps from previous years.

ariasuni:
Fix #833 

Since #509, the format used for the date of the current year was not taking locale into account.

It does now, with a simpler and shorter code, and a test.


ogham:
Excellent work once again, thanks.

PR #841, opened 2021-04-09
Fix --git deducing ignored state in unintuitive way

### LINKED ISSUE #840: Don’t show a directory as ignored by git because something is ignored in it
Right now, when using `--git`, if a directory has an ignored file somewhere under it, it shows `-I` in the git column.

While it’s useful to know something had been added or modified in a directory, it’s much more confusing to see a directory marked as ignored while it’s not, only because something is ignored under it. We could think of it as the fact that, from a user perspective, adding a file “modifies” its parent directories, but ignoring a directory works the other way around and ignores everything under it.

So, I believe exa should only display the ignored state of a directory itself if it is itself ignored, or if a parent directory is ignored.

(The idea for this bug report comes from #578)

ariasuni:
Fix #840 

It makes sense to consider that a directory has the aggregated status of all the files under it.

The exception is that for the ignored status, it’s more useful and intuitive to consider that it applies to everything under it.

- A directory containing an ignored file is no longer considered ignored
- A file inside an ignored directory is now considered ignored

### DIFF
--- a/src/fs/feature/git.rs
+++ b/src/fs/feature/git.rs
@@ -252,15 +252,21 @@ impl Git {
             .unwrap_or_default()
     }
 
-    /// Get the combined status for all the files whose paths begin with the
-    /// path that gets passed in. This is used for getting the status of
-    /// directories, which don’t really have an ‘official’ status.
+    /// Get the combined, user-facing status of a file or directory.
+    /// Statuses are aggregating (for example, a directory is considered
+    /// modified if any file under it has the status modified), except
+    /// for ignored which applies to files under (for example, a file is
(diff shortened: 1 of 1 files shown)


ogham:
You are on a roll!

This makes sense to me, too — I've also been confused about whether it's a directory or a file that is being ignored. However, I didn't want to dive back into the ignore code because I assumed it would be tricky to fix. Looks like you found a nice and simple way to do it though.

ariasuni:
Ah damn you’re too fast! I didn’t finished the PR — it was wrong and it lacks test. :sob:

ogham:
Haha, whoops!

Just submit another one when it's ready, no problem.

PR #842, opened 2021-04-09
Fix deducing git ignored state for files too, not only directories

### LINKED ISSUE #656: [Enhancement] Make the `I` indicator in the git column only apply to the gitignore matches
Imagine you have a git repo with the following structure

```
foo
|
\- bar
|  |
|  \- baz
|     |
|     \- project.iml
|
\- baz
```
and a gitignore that just contains `*.iml`.

At the moment, `exa -al --git` in the `foo/bar/baz` directory will produce the following output
```
<perms> <size> <username> <date modified>  -I  project.iml
```
which is expected. However, `exa -al --git` also produces the `-I` for the `bar` folder when run in the `foo` directory, and the `baz` directory when run in the `foo/bar` directory.

I would have expected this feature to only apply to direct matches. If the intent was to indicate that a directory contains an ignored file, that's understandable. However, it makes the meaning of `-I` ambiguous

ariasuni:
Fix #656 (I knew I had the idea for this PR from somewhere!!).

Follow-up of #841.

- The aggregation logic was only implemented for directories, since it doesn’t make sense for files. However, when I added the logic that a directory is considered ignored if a parent directory is, I forgot to implement it for files too.
- Added tests to check that a non-ignored parent directory of an ignored content isn’t flagged as ignored, and that the content of an ignored directory is flagged as ignored too.

This should be good this time!! Next time, I should probably use the draft PR feature when I know I need to add or fix stuff…


ogham:
Thanks again! This is golden.

ariasuni:
Happy to help!

PR #846, opened 2021-04-14
Add support for querying timezone on Android

skyline75489:


skyline75489:
OK the code does *not* work. The compiler complains:

```text
error[E0283]: type annotations needed
   --> src/output/table.rs:324:20
    |
324 |             return CompiledData::parse(output.stdout);
    |                    ^^^^^^^^^^^^^^^^^^^ cannot infer type
    |
    = note: cannot satisfy `_: CompiledData`
    = note: required by `zoneinfo_compiled::CompiledData::parse`

error: aborting due to previous error; 1 warning emitted
```

But I honestly don't know what's wrong with it.

src/output/table.rs
```diff
@@ -316,7 +316,15 @@ fn determine_time_zone() -> TZResult<TimeZone> {
             }
         })
     } else {
-        TimeZone::from_file("/etc/localtime")
+        if env::consts::OS == "android" {
```
ariasuni, line 319: I suppose it’s a good practice to include the “documentation” for this change.

```suggestion
        /// https://android.googlesource.com/platform/bionic/+/master/libc/tzcode/bionic.cpp
        if env::consts::OS == "android" {
```

ariasuni (changes requested):

ariasuni:
Well actually your approach is wrong. `persist.sys.timezone` doesn’t contain the data to parse but I guess the filename relative to the directory containing the `TZDATA`, and Bionic’s source code seems to also take into account the «first format» described at https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html.

I’m really starting to think we should stop trying to replace the `libc` and use standard C function so that compatibility with OS isn’t such a nightmare of trying to understand what such or such libc are actually doing.

skyline75489:
Closed in favor of #867

PR #858, opened 2021-04-26
Add F# icons

Prunkles:
Used icon `nf-dev-fsharp` for `.fs`, `.fsi` and `.fsx` extensions


ariasuni:
Wikipedia says that F# file extensions are «.fs, .fsi, .fsx, .fsscript». I don’t know anything about F#, so do you think it’s relevant to add this extension too, and if yes can you add it? Thanks :)

Prunkles:
I've been writing in F# for over a year but have never seen this extension. `.fsx` is now used for scripts, so I think `.fsscript` is legacy

ariasuni:
A quick web search let me think you’re right.

Issue #757, opened 2020-10-26
Website information incorrect?


sstainba:
The website "the.exa.website" indicates there are packages on apt for more than just Ubuntu gorilla, but when searching the package repo, only the gorilla package is found.

The manual install steps provide the man page link as "https://raw.githubusercontent.com/ogham/exa/master/contrib/man/exa.1" which is invalid.

ohmyhub:
The manual installation instructions found at [https://the.exa.website/install/linux#manual](https://the.exa.website/install/linux#manual) has several broken links.

Corrected location of the man page:
[https://github.com/ogham/exa/blob/master/man/exa.1.md](https://github.com/ogham/exa/blob/master/man/exa.1.md)

Corrected location of shell completion files:
bash - [https://github.com/ogham/exa/blob/master/completions/completions.bash](https://github.com/ogham/exa/blob/master/completions/completions.bash)
zsh - [https://github.com/ogham/exa/blob/master/completions/completions.zsh]( https://github.com/ogham/exa/blob/master/completions/completions.zsh)
fish - [https://github.com/ogham/exa/blob/master/completions/completions.fish](https://github.com/ogham/exa/blob/master/completions/completions.fish)

ariasuni:
Also, exa’s website doesn’t link to the splitted [man/exa_colors.5.md](https://github.com/ogham/exa/blob/v0.10.1/man/exa_colors.5.md). Maybe links should include the version so it doesn’t break during a development phase or point to new features not yet stable, and updating them is another thing to update on the website for each new release.

ariasuni:
The links to exa’s releases in the [Linux manual installation instructions](https://the.exa.website/install/linux#manual) are wrong, too (refer to 0.10.0 instead of 0.10.1).

KOGI:
Any movement on this? These links are still broken

PR #860, opened 2021-04-26
Rename completion scripts according to each shell conventions

### LINKED ISSUE #451: Split the completions directory and rename scripts according to conventions of each shell
First of all, thanks a lot for this great tool!

Would it be possible to split the [`/contrib`](/ogham/exa/tree/master/contrib) subdirectory for each shell? This way, I could add the completions script easily via [antigen](/zsh-users/antigen): `antigen bundle ogham/exa contrib/completions/zsh`.

Thanks again!

ariasuni:
Fix #451 

```
renamed:         completions/completions.bash -> completions/bash/exa
renamed:         completions/completions.fish -> completions/fish/exa.fish
remaned:         completions/completions.zsh -> completions/zsh/_exa
```


ariasuni:
I’ll merge this now, because the website’s links are already broken and this is tracked at #757 anyway.

timbedard:
Heads up: This breaks Homebrew `--HEAD` installation. We'll need to add a check of some sort to [the formula](https://github.com/Homebrew/homebrew-core/blob/11d3369762a7eccf2349eefc732ffe183cb0ba13/Formula/exa.rb) to determine where to find the completions.

ariasuni:
@timbedard Should I open an issue there? I wouldn’t be comfortable making a PR I couldn’t test, and the changes are trivial so…

timbedard:
@ariasuni I usually pull up [this guide](https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request) when I need to open a PR for Homebrew. It's not hard to run the existing tests and it's outlined in that guide.

ariasuni:
Well to be honest I don’t how `--HEAD` works and how to not break it for users of the stable version, and it certainly looks like someone who uses brew should open the issues since it requests output from `brew` and warns about not filling the template. And I couldn’t make brew works on my (Linux) system last time I tried and it was a mess so, eh.

timbedard:
Gotcha. I'll try to give it a go sometime this week when I have time.

PR #862, opened 2021-04-30
Cleanup clippy warnings

cgzones:


ariasuni (approved):
Thanks, it was very annoying but I couldn’t be bothered to look at the warnings.

PR #869, opened 2021-05-11
Make sort more consistent accross uses and flags

### LINKED ISSUE #863: --reverse does not reverse in the case of a tie
Exa version v0.10.1 [+git], NixOS.

If two items are the same in the sort order, `--reverse` will not reverse their order. The easiest demonstration is to create three files at the same time

```bash
touch file1 file2 file3
```

and sort by newest
```bash
exa -snew
```
```
file1  file3  file2
```
This is "directory order" (`ls -U`) so it may be different from this example. If we try to reverse this order...
```bash
exa -r -snew
```
```
file1  file3  file2
```
we still get the directory order!  think this behaviour is surprising, and can be particularity vexacious when there are only some ties.

In `ls` (GNU, busybox, and toybox at least) the tie appears to be broken by filename:
```bash
ls -t
```
```
file1 file2 file3
```
and
```bash
ls -rt
```
```
file3 file2 file1
```
to me this is sensible behaviour, but in any case I think the `--reverse` flag should always produce the opposite order of files within a directory to the same command without the `--reverse` flag.

ariasuni:
Fix #863

In some cases, files were sorted twice, which in most cases is harmless.
But when two files are determined equal, their order is unchanged (because the sort is stable).
With `--reverse`, equal elements order, which wasn’t changed by the sort, were reversed twice (unchanged).

Also, command lines arguments are now sorted by their path.

---

I still need to check the xtests to see if anything’s wrong with my changes, and update the tests that needs it.

### DIFF
--- a/src/fs/filter.rs
+++ b/src/fs/filter.rs
@@ -88,11 +88,11 @@ impl FileFilter {
     }
 
     /// Sort the files in the given vector based on the sort field option.
-    pub fn sort_files<'a, F>(&self, files: &mut Vec<F>)
+    pub fn sort_files<'a, F>(&self, files: &mut Vec<F>, maybe_different_parents: bool)
     where F: AsRef<File<'a>>
     {
         files.sort_by(|a, b| {
-            self.sort_field.compare_files(a.as_ref(), b.as_ref())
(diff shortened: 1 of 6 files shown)


micwoj92:
This branch has conflicts that must be resolved

Issue #568, opened 2019-08-10
--icons flag not present in output of exa --help


andradei:
Makes me think other flags might also not be present in `exa --help` and `man exa`.

`--icons` is missing for sure.

ariasuni:
It is also not present in the README, and there should be a mention of what font you need to see the icons.

j-tai:
This was already fixed in commit `4d9c6eb`, which will be in the next release.

ariasuni:
@j-tai You should put commit hash directly (without `) so that GitHub can link to it, ex: 4d9c6eb

So, the issue is that only the `README` is missing this information.

BarbzYHOOL:
also the font is not working in 0.9

SuperSandro2000:
This should really be added because it makes discovering that option really hard.

ariasuni:
@BarbzYHOOL You need to use [Nerd Fonts](https://www.nerdfonts.com/). I’ll improve the documentation so that everything is clear about `--icons`.

PR #880, opened 2021-05-21
Add Perl module and test script icons

mjgardner:


ariasuni:
Wikipedia lists as Perl extensions: `plx`, `pl`, `pm`, `xs`, `t`, `pod`. Right now, only `pl` is taken into account, and you’re proposing to add `pm` and `t`. What about the other extensions, are they legacy or too specific?

`plx` seems legacy, `pod` is documentation so another icon is probably more appropriate, and `xs` seems like something that could use the same icon as other Perl files.

gek169:
this is an ls replacement, why do we have icons

ariasuni:
Because `exa` doesn’t try to limit itself to what `ls` does, and that a lot of people find icons nice and/or useful, and it’s optional so nobody has to use it if they don’t like them. File managers have icons, I don’t see why a file lister can have them too.

gek169:
so what this is using w3m-img or something to display images in the terminal?

ariasuni:
It’s characters that display correctly if you install a font from [Nerd Fonts](https://www.nerdfonts.com/). I need to document that properly, [as I wrote in another issue](https://github.com/ogham/exa/issues/568#issuecomment-753356833).

gek169:
Oh that's neat

ariasuni:
Hi @mjgardner, did you see [my comment](https://github.com/ogham/exa/pull/880#issuecomment-869178331)?

PR #897, opened 2021-06-22
Call `{symlink_,}metadata` only when really needed

oxalica:
As we know `lstat` is very slow when metadata is not in cache.
When simply listing files and color is diabled (stdout is piped), all information from `read_dir` is enough. We can reduce tons of `lstat`/`statx` in case of lots of files in a single directory, and speeds up about 10x when listing a directory with 25k+ files.

```
$ echo 3 | sudo tee /proc/sys/vm/drop_caches
3

$ time target/release/exa /nix/store | wc -l # <- after this patch
26227
target/release/exa /nix/store  0.07s user 0.03s system 69% cpu 0.147 total
wc -l  0.01s user 0.01s system 11% cpu 0.147 total

$ echo 3 | sudo tee /proc/sys/vm/drop_caches
3

$ time exa /nix/store | wc -l # <- before this patch
26227
exa /nix/store  0.08s user 0.29s system 27% cpu 1.367 total
wc -l  0.00s user 0.02s system 1% cpu 1.367 total

$ echo 3 | sudo tee /proc/sys/vm/drop_caches
3

$ time ls /nix/store | wc -l # <- for comparison 
26227
ls /nix/store  0.04s user 0.01s system 60% cpu 0.084 total
wc -l  0.00s user 0.00s system 4% cpu 0.084 total
```

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -1,5 +1,6 @@
 //! Files, and methods and fields to access their metadata.
 
+use std::fs::{FileType, Metadata};
 use std::io;
 use std::os::unix::fs::{FileTypeExt, MetadataExt, PermissionsExt};
 use std::path::{Path, PathBuf};
@@ -40,12 +41,19 @@ pub struct File<'dir> {
     /// path (following a symlink).
     pub path: PathBuf,
 
+    /// The type of this file. None for directories `.` and `..`.
+    file_type: Option<FileType>,
+
     /// A cached `metadata` (`stat`) call for this file.
     ///
     /// This too is queried multiple times, and is *not* cached by the OS, as
     /// it could easily change between invocations — but exa is so short-lived
     /// it’s better to just cache it.
-    pub metadata: std::fs::Metadata,
+    /// It may be None if the call to symlink_metadata fails.
+    ///
+    /// In case of metadata is not needed, like simple listing without color,
+    /// it would be None.
+    metadata: Option<Metadata>,
 
     /// A reference to the directory that contains this file, if any.
     ///
@@ -66,42 +74,66 @@ pub struct File<'dir> {
 }
 
 impl<'dir> File<'dir> {
-    pub fn from_args<PD, FN>(path: PathBuf, parent_dir: PD, filename: FN) -> io::Result<File<'dir>>
+    pub fn from_args<PD, FN>(
+        path: PathBuf,
+        parent_dir: PD,
+        filename: FN,
+        file_type: Option<FileType>,
+        needs_metadata: bool,
+    ) -> io::Result<File<'dir>>
     where PD: Into<Option<&'dir Dir>>,
           FN: Into<Option<String>>
     {
         let parent_dir = parent_dir.into();
         let name       = filename.into().unwrap_or_else(|| File::filename(&path));
         let ext        = File::ext(&path);
-
-        debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
         let is_all_all = false;
 
-        Ok(File { name, ext, path, metadata, parent_dir, is_all_all })
+        // For files other than `.` and `..`, `file_type` must be present.
+        // `file_type` is used to determine whether it is a directory when parsing command line arguments.
+        let (file_type, metadata) = match (needs_metadata, file_type) {
+            (false, Some(file_type)) => (Some(file_type), None),
+            (true, _) | (false, None) => {
+                debug!("Statting file {:?}", &path);
+                let metadata = std::fs::symlink_metadata(&path)?;
+                (Some(metadata.file_type()), Some(metadata))
+            }
+        };
+
+        Ok(File { name, ext, path, file_type, metadata, parent_dir, is_all_all })
     }
 
-    pub fn new_aa_current(parent_dir: &'dir Dir) -> io::Result<File<'dir>> {
+    pub fn new_aa_current(parent_dir: &'dir Dir, needs_metadata: bool) -> io::Result<File<'dir>> {
         let path       = parent_dir.path.clone();
         let ext        = File::ext(&path);
 
-        debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
+        let (file_type, metadata) = if needs_metadata {
+            debug!("Statting file {:?}", &path);
+            let metadata = std::fs::symlink_metadata(&path)?;
+            (Some(metadata.file_type()), Some(metadata))
+        } else {
+            (None, None)
+        };
         let is_all_all = true;
         let parent_dir = Some(parent_dir);
 
-        Ok(File { path, parent_dir, metadata, ext, name: ".".into(), is_all_all })
+        Ok(File { path, parent_dir, file_type, metadata, ext, name: ".".into(), is_all_all })
     }
 
-    pub fn new_aa_parent(path: PathBuf, parent_dir: &'dir Dir) -> io::Result<File<'dir>> {
+    pub fn new_aa_parent(path: PathBuf, parent_dir: &'dir Dir, needs_metadata: bool) -> io::Result<File<'dir>> {
         let ext        = File::ext(&path);
 
-        debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
+        let (file_type, metadata) = if needs_metadata {
+            debug!("Statting file {:?}", &path);
+            let metadata = std::fs::symlink_metadata(&path)?;
+            (Some(metadata.file_type()), Some(metadata))
+        } else {
+            (None, None)
+        };
         let is_all_all = true;
         let parent_dir = Some(parent_dir);
 
-        Ok(File { path, parent_dir, metadata, ext, name: "..".into(), is_all_all })
+        Ok(File { path, parent_dir, file_type, metadata, ext, name: "..".into(), is_all_all })
     }
 
     /// A file’s name is derived from its string. This needs to handle directories
@@ -134,9 +166,20 @@ impl<'dir> File<'dir> {
             .to_ascii_lowercase())
     }
 
+    // This method can only be called if metadata is initialized (`needs_metadata` is true for
+    // constructor), or it will panic.
+    fn metadata(&self) -> &Metadata {
+        self.metadata.as_ref().expect("metadata not initialized")
+    }
+
+    /// The file size reported by metadata.
+    pub fn file_len(&self) -> u64 {
+        self.metadata().len()
+    }
+
     /// Whether this file is a directory on the filesystem.
     pub fn is_directory(&self) -> bool {
-        self.metadata.is_dir()
+        self.file_type.map_or(true, |t| t.is_dir())
     }
 
     /// Whether this file is a directory, or a symlink pointing to a directory.
@@ -168,7 +211,7 @@ impl<'dir> File<'dir> {
     /// Whether this file is a regular file on the filesystem — that is, not a
     /// directory, a link, or anything else treated specially.
     pub fn is_file(&self) -> bool {
-        self.metadata.is_file()
+        self.file_type.map_or(false, |t| t.is_file())
     }
 
     /// Whether this file is both a regular file *and* executable for the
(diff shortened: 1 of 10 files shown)


src/fs/file.rs
```diff
@@ -40,12 +42,16 @@ pub struct File<'dir> {
     /// path (following a symlink).
     pub path: PathBuf,
 
+    /// The type of this file. None for directories `.` and `..`.
+    file_type: Option<FileType>,
```
ariasuni, line 46: Did you check if it made a difference? Calling `file_type()` one time vs calling the various functions doesn’t poll the filesystem each time and doesn’t make a complex operation, [see there](https://github.com/rust-lang/rust/blob/543ab9964065dc2b4e47d7bff5e83c84519f265e/library/std/src/sys/unix/fs.rs#L270). Adding complexity isn’t worth it if this is optimized away by the compiler or if the different isn’t measurable.

src/fs/file.rs
```diff
@@ -40,12 +42,16 @@ pub struct File<'dir> {
...
+
     /// A cached `metadata` (`stat`) call for this file.
     ///
     /// This too is queried multiple times, and is *not* cached by the OS, as
     /// it could easily change between invocations — but exa is so short-lived
     /// it’s better to just cache it.
-    pub metadata: std::fs::Metadata,
+    /// It may be None if the call to symlink_metadata fails.
+    metadata: OnceCell<Option<Metadata>>,
```
ariasuni, line 54: I’m not sure why you’re using `OnceCell` here (to be honest I’ve checked it and didn’t really understood what it was about), could you explain why you used it?

ariasuni (commented):
Thanks for your work, this seems interesting!

Could you please split the changes related to metadata and `FileType` and the ones related to `LinkStyle` into two commits? It would make it much easier to review.

src/fs/file.rs
oxalica: It's not for lazy. It's just that we can't construct it.
Normally `FileType` is from `DirEntry` when reading directories. As I said in comment, this field will be always `Some` for real `DirEntry` from `read_dir`, except for `.` and `..`, which are not returned by `read_dir` and are created as dummy instance with [these two functions](https://github.com/ogham/exa/pull/897/files#diff-c464eda590357178f19a588f108f7a09241fb6125aeaf8ea5acda2a501d06112L84-L105) .
 Since there is no real syscall, we cannot construct a `FileType`. I use `None` for these dummy `File` and make getter functions, like `is_directory`, correctly handle this case.

src/fs/file.rs
oxalica: In case of colorful output, we will determine if a file `is_executable_file`, which requires metadata (for permission informatino). But if not, we don't need `Metadata` at all. So I made it a lazy field: only invoke syscall if really needed.
We construct `File`s during scanning, not rendering. So it is not easy to determine whether this field is needed when constructing `File`.

BTW: `unsync::OnceCell` should be more efficient. But we use multi-threading and pass `&File` to other threads. Therefore, only `sync::OnceCell` can be used, though there should be only one thread accessing it at a time.

ariasuni: Well, if I understand correctly, the only case where `metadata` is not initialized should be with `.` and `..`. Could the usage of `OnceCell` be avoided by checking `is_all_all` in the `metadata()` function?

src/fs/file.rs
ariasuni: But, couldn’t e.g. `self.file_type.map_or(false, |t| t.is_file())` be replaced by `self.metadata.file_type().map_or(false, |t| t.is_file())`, or I am missing something?

Sorry if it’s bothering you, but since it’s both a complicated part of the code and an important one, I have difficulty understanding everything that’s going on and I’d like not complicate it further if not strictly necessary. Your contribution is much appreciated. :slightly_smiling_face:

src/fs/file.rs
oxalica: > the only case where metadata is not initialized should be with . and ..

No. In my benchmark in the PR description (used as `exa /nix/store | wc -l`), there is no `metadata` being called at all because color is disabled for pipe. That's why it's 10x faster.
`.` and `..` are already not calling `metadata` on master. That's not the key point.

src/fs/file.rs
oxalica: `metadata` is expensive. This PR is just to elide them at all by use `file_type` as much possible, which is free from `DirEntry` without any syscall.
See also https://github.com/ogham/exa/pull/897#discussion_r659385287

ariasuni: Ok, thank you for the explanation!

oxalica:
To be clear, `exa` relies on information which can only be got from `metadata` when:
- With `-l`: Need to print permission, size, etc.
- Without `-l` but with color: Need to render special color for executable file (permission).

Other operation like `is_link`, `is_dir`, `is_socket`, etc, can be determined by `FileType`, which is free from `DirEntry`.

So in case of "without `-l`, no color", eg. `exa /some/path | some_other_program`, all invocations to `metadata` can be elided.
That's what I did in this PR.

ariasuni:
I think one problem with your current approach is that can since metadata is not fetched at the creation of the file, all functions should return an error and it would change a lot of functions.

Isn’t possible to make `metadata` an `Option<Metadata>` and check if metadata will be needed earlier in the code? See my incomplete patch below. It misses checking for sort flags needing metadata, and adapting the functions using `file_type()` and `.metadata()`)`:

<details><summary>code proposition</summary>

```patch
diff --git a/src/fs/dir.rs b/src/fs/dir.rs
index 1a6ce01..ea831d6 100644
--- a/src/fs/dir.rs
+++ b/src/fs/dir.rs
@@ -47,12 +47,13 @@ impl Dir {
 
     /// Produce an iterator of IO results of trying to read all the files in
     /// this directory.
-    pub fn files<'dir, 'ig>(&'dir self, dots: DotFilter, git: Option<&'ig GitCache>, git_ignoring: bool) -> Files<'dir, 'ig> {
+    pub fn files<'dir, 'ig>(&'dir self, dots: DotFilter, git: Option<&'ig GitCache>, git_ignoring: bool, needs_metadata: bool) -> Files<'dir, 'ig> {
         Files {
             inner:     self.contents.iter(),
             dir:       self,
             dotfiles:  dots.shows_dotfiles(),
             dots:      dots.dots(),
+            needs_metadata,
             git,
             git_ignoring,
         }
@@ -86,6 +87,8 @@ pub struct Files<'dir, 'ig> {
     /// any files have been listed.
     dots: DotsNext,
 
+    needs_metadata: bool,
+
     git: Option<&'ig GitCache>,
 
     git_ignoring: bool,
@@ -118,7 +121,7 @@ impl<'dir, 'ig> Files<'dir, 'ig> {
                     }
                 }
 
-                return Some(File::from_args(path.clone(), self.dir, filename)
+                return Some(File::from_args(path.clone(), self.dir, filename, self.needs_metadata)
                                  .map_err(|e| (path.clone(), e)))
             }
 
@@ -148,13 +151,13 @@ impl<'dir, 'ig> Iterator for Files<'dir, 'ig> {
         match self.dots {
             DotsNext::Dot => {
                 self.dots = DotsNext::DotDot;
-                Some(File::new_aa_current(self.dir)
+                Some(File::new_aa_current(self.dir, self.needs_metadata)
                           .map_err(|e| (Path::new(".").to_path_buf(), e)))
             }
 
             DotsNext::DotDot => {
                 self.dots = DotsNext::Files;
-                Some(File::new_aa_parent(self.parent(), self.dir)
+                Some(File::new_aa_parent(self.parent(), self.dir, self.needs_metadata)
                           .map_err(|e| (self.parent(), e)))
             }
 
diff --git a/src/fs/file.rs b/src/fs/file.rs
index ea83f08..db624a7 100644
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -45,7 +45,7 @@ pub struct File<'dir> {
     /// This too is queried multiple times, and is *not* cached by the OS, as
     /// it could easily change between invocations — but exa is so short-lived
     /// it’s better to just cache it.
-    pub metadata: std::fs::Metadata,
+    pub metadata: Option<std::fs::Metadata>,
 
     /// A reference to the directory that contains this file, if any.
     ///
@@ -66,7 +66,7 @@ pub struct File<'dir> {
 }
 
 impl<'dir> File<'dir> {
-    pub fn from_args<PD, FN>(path: PathBuf, parent_dir: PD, filename: FN) -> io::Result<File<'dir>>
+    pub fn from_args<PD, FN>(path: PathBuf, parent_dir: PD, filename: FN, needs_metadata: bool) -> io::Result<File<'dir>>
     where PD: Into<Option<&'dir Dir>>,
           FN: Into<Option<String>>
     {
@@ -75,29 +75,29 @@ impl<'dir> File<'dir> {
         let ext        = File::ext(&path);
 
         debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
+        let metadata   = if needs_metadata { Some(std::fs::symlink_metadata(&path)?) } else { None };
         let is_all_all = false;
 
         Ok(File { name, ext, path, metadata, parent_dir, is_all_all })
     }
 
-    pub fn new_aa_current(parent_dir: &'dir Dir) -> io::Result<File<'dir>> {
+    pub fn new_aa_current(parent_dir: &'dir Dir, needs_metadata: bool) -> io::Result<File<'dir>> {
         let path       = parent_dir.path.clone();
         let ext        = File::ext(&path);
 
         debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
+        let metadata   = if needs_metadata { Some(std::fs::symlink_metadata(&path)?) } else { None };
         let is_all_all = true;
         let parent_dir = Some(parent_dir);
 
         Ok(File { path, parent_dir, metadata, ext, name: ".".into(), is_all_all })
     }
 
-    pub fn new_aa_parent(path: PathBuf, parent_dir: &'dir Dir) -> io::Result<File<'dir>> {
+    pub fn new_aa_parent(path: PathBuf, parent_dir: &'dir Dir, needs_metadata: bool) -> io::Result<File<'dir>> {
         let ext        = File::ext(&path);
 
         debug!("Statting file {:?}", &path);
-        let metadata   = std::fs::symlink_metadata(&path)?;
+        let metadata   = if needs_metadata { Some(std::fs::symlink_metadata(&path)?) } else { None };
         let is_all_all = true;
         let parent_dir = Some(parent_dir);
 
@@ -253,7 +253,7 @@ impl<'dir> File<'dir> {
             Ok(metadata) => {
                 let ext  = File::ext(&path);
                 let name = File::filename(&path);
-                let file = File { parent_dir: None, path, ext, metadata, name, is_all_all: false };
+                let file = File { parent_dir: None, path, ext, metadata: Some(metadata), name, is_all_all: false };
                 FileTarget::Ok(Box::new(file))
             }
             Err(e) => {
diff --git a/src/main.rs b/src/main.rs
index 6730993..1810a19 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -168,7 +168,7 @@ impl<'args> Exa<'args> {
         let mut exit_status = 0;
 
         for file_path in &self.input_paths {
-            match File::from_args(PathBuf::from(file_path), None, None) {
+            match File::from_args(PathBuf::from(file_path), None, None, self.needs_metadata()) {
                 Err(e) => {
                     exit_status = 2;
                     writeln!(io::stderr(), "{:?}: {}", file_path, e)?;
@@ -221,7 +221,7 @@ impl<'args> Exa<'args> {
 
             let mut children = Vec::new();
             let git_ignore = self.options.filter.git_ignore == GitIgnore::CheckAndIgnore;
-            for file in dir.files(self.options.filter.dot_filter, self.git.as_ref(), git_ignore) {
+            for file in dir.files(self.options.filter.dot_filter, self.git.as_ref(), git_ignore, self.needs_metadata()) {
                 match file {
                     Ok(file)        => children.push(file),
                     Err((path, e))  => writeln!(io::stderr(), "[{}: {}]", path.display(), e)?,
@@ -316,6 +316,13 @@ impl<'args> Exa<'args> {
             }
         }
     }
+
+    fn needs_metadata(&self) -> bool {
+        match self.options.view.mode {
+            Mode::Details(_) | Mode::GridDetails(_) => true,
+            _ => self.theme.ui.colourful,
+        }
+    }
 }
 
 
diff --git a/src/output/details.rs b/src/output/details.rs
index 9dca7d4..38be291 100644
--- a/src/output/details.rs
+++ b/src/output/details.rs
@@ -296,7 +296,7 @@ impl<'a> Render<'a> {
             rows.push(row);
 
             if let Some(ref dir) = egg.dir {
-                for file_to_add in dir.files(self.filter.dot_filter, self.git, self.git_ignoring) {
+                for file_to_add in dir.files(self.filter.dot_filter, self.git, self.git_ignoring, true) {
                     match file_to_add {
                         Ok(f) => {
                             files.push(f);
```
</details>

oxalica:
@ariasuni 
> Isn’t possible to make metadata an Option<Metadata> and check if metadata will be needed earlier in the code? See my incomplete patch below.

Thanks for the advice. I refactored it and make early decision now, which doesn't need `OnceCell` anymore.

> Could you please split the changes related to metadata and FileType and the ones related to LinkStyle into two commits? It would make it much easier to review.

Splitted.

ariasuni:
If you ever have the time and want to do work on this again, there is an [active fork called eza](https://github.com/eza-community/eza), I’m sure your contribution would be appreciated. I would totally understand if you couldn’t or didn’t want, though. Thanks anyway! and sorry to have been unable to review your work

PR #899, opened 2021-06-24
Fix fish completion for `-H` option

kidonng:
> `-h`, `--header`
> : Add a header row to each column.
> 
> `-H`, `--links`
> : List each file’s number of hard links.


ariasuni:
Good catch, thanks!

PR #903, opened 2021-07-02
fix: more JPG extensions

xxkfqz:
Added:
* `.jfi`
* `.jfif`
* `.jif`
* `.jpe`

### DIFF
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -35,10 +35,10 @@ impl FileExtensions {
 
     fn is_image(&self, file: &File<'_>) -> bool {
         file.extension_is_one_of( &[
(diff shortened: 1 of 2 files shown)


gek169:
.jif?

ariasuni:
I don’t think I’ve ever seen any of these extensions, except `.jpe` on Twitter, and I can’t really find accurate information about it apart from the fact that it’s sometimes is/was used. But I guess it doesn’t hurt to support them, at least until maybe someday it’s used for a different file format.

@gek169 You can find information about it [on Wikipedia](https://en.wikipedia.org/wiki/JPEG#JPEG_filename_extensions). Please take the time to search for information by yourself before asking here. Also more detailed answers makes it easier to answer (and to give a good answer).

PR #908, opened 2021-07-06
Don’t show icons when piped (also add `--icons={always,auto,never}`)

### LINKED ISSUE #691: No way to disable --icons on piping
exa --icons sends icons across the pipe. If there was an --icons=WHEN option, icons could be disabled when being piped similarly to --color

ariasuni:
Fix #691 

what needs to be done:

- [ ] Add tests
- [ ] Decide if we want the same behavior for `--classify`/`-F` as well.


meryacine:
Hello,
Is this still being worked on?

Issue #627, opened 2020-01-26
Add Podfile to Immediate files


pro-sumer:
_[CocoaPods](https://cocoapods.org) is a dependency manager for Swift and Objective-C Cocoa projects. The [Podfile](https://guides.cocoapods.org/using/the-podfile.html) is a specification that describes the dependencies of the targets of one or more Xcode projects. The file should simply be named `Podfile`._

Please add `Podfile` to the "Immediate" class.

PR #888, opened 2021-06-06
fix:  Compilation failed on aarch64

neepoo:
I fixed [this issue](https://github.com/ogham/exa/issues/887) with conditional compilation and its compile error hint.

But I'm new to rust, I can compile successfully now under linux with x86_64 and aarch64 architecture, I don't know if it will affect other operating systems or architectures.

Thank you, code reviewer!

PR #855, opened 2021-04-25
Add option to show security attribute and improve extended support

cgzones:
Add a command line option `-Z/--context` to show the security context of objects, similar to ls(1).

Show the actual extended attribute values on `-@/--extended`, instead of just their length.

In case of a symlink, show the extended attributes of the symlink itself, not the target.
This matches the behavior of ls(1) and is more intuitive.

TODO:
   * test on `macos` (*should* compile)

In the future one might want to add support for SMACK security labels (`"security.SMACK64"`).

Closes: #254 #613

exa

exa2

PR #909, opened 2021-07-08
Use safer method of pointer casting when getting extended attributes.

illogic-al:
Hey. I'm new to rust (and using exa which is _amazing_). I ran clippy and it suggested these changes to the unsafe pointer casting that's happening. 

I ran the cargo tests and they all but, but I noticed I didn't see anything explicitly in the list about xattrs when running. 

I'd be willing to *try* write some tests ,and test on macOS and linux, if needed to verify these changes are OK.

### DIFF
--- a/src/fs/feature/xattr.rs
+++ b/src/fs/feature/xattr.rs
@@ -167,7 +167,7 @@ mod lister {
             unsafe {
                 listxattr(
                     c_path.as_ptr(),
-                    buf.as_mut_ptr() as *mut c_char,
+                    buf.as_mut_ptr().cast::<i8>(),
                     bufsize as size_t,
                     self.c_flags,
                 )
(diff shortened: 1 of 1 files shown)


ariasuni:
Hi! I’m very glad you enjoy exa. I hope you’re enjoying Rust too!

There’s already #855 where someone modifies [this code](https://github.com/ogham/exa/pull/855/files#diff-758a32aff9f7e4f5f9337b84e24da3e3b77c8e9ce6518304f3d27f0b26913132R242-R243). Also, unfortunately, this will not work for some architectures this is being worked on in #888. (yeah I didn’t have much time to review and merge these)

So I’ll close this issue. Sorry, you’re quite unlucky with your PR ^^’

If you want to work on something else, I could help you find easy PR or understand the code. For example, you could try to fix #627.

illogic-al:
> Nah, closing is fine if it's not a good idea. I would like to try with that other commit you mentioned but ... I don't get it. How would I add a podfile?

Well I see that that ticket is closed :-D
I also see how podfile support was added, but why was an 'immediate' file referenced, no such file was modified?

PR #867, opened 2021-05-09
Replace datetime and zoneinfo_compiled crates with chrono

ariasuni:
After hours of pain trying to retro-engineer several libc including glibc, musl and Bionic (Android), I’ve come to the conclusion that it’s the wrong approach. But I couldn’t understand the mess that’s time and timezone handling in C, much less investigate every OS we would like exa to support.

Fortunately, [`chrono`](https://crates.io/crates/chrono) already provides us a nice Rust layer on top of all the nasty C platform-dependent code. Apart from Android, chrono also supports Windows.

- Improve compatibility with other OSes, timezone are handled for us
- Reduce significantly code handling and rendering time

Fix #620 
Fix #712 
Fix #856
Fix #1047 

Close #846
Close #857

PR #911, opened 2021-07-12
Fix time zone handling with TZ=:/absolute/path

mrkline:
determine_time_zone() assumes that a `TZ` environment variable that
starts with `:` is a relative path, so `TZ=:/absolute/path` tries
(and fails) to load zone info from `/usr/share/zoneinfo//absolute/path`.

The GNU libc manual suggests that both `:relative/path`
(mapping to `/usr/share/zoneinfo/relative/path`)
and `:/absolute/path` should work.

### DIFF
--- a/src/output/table.rs
+++ b/src/output/table.rs
@@ -303,23 +303,27 @@ impl Environment {
 }
 
 fn determine_time_zone() -> TZResult<TimeZone> {
+    // From the GNU libc manual,
+    // https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html:
+    //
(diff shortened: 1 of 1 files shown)


ariasuni:
Ah, I’m sorry! I already did the job there: #857, but I didn’t merge it. (Also, the `:` at the start is removed by both glibc and musl, it’s the `/` at the start of the path that tells if it’s an absolute path or not)

I thought about it and then wrote #867 which might fix the problem in a better way, but since it’s a bigger change, I’d prefer to know if exa’s author agree with it.

Feel free to comment on #857 if you think it could be improved, though.

Issue #916, opened 2021-07-24
Include "*.mk" for makefile icon


cory02:
Currently only files named "Makefile" are treated with the special icon.  Please consider including files with extension "mk".

ariasuni:
This has been done :slightly_smiling_face:, thanks @joshuagawley.

PR #918, opened 2021-07-29
Display Makefile icon for .mk files

joshuagawley:
Displays the Makefile icon for .mk files, as desired by #916 .


ariasuni:
Well, I don’t agree with the icon for `Makefile`, there are other implementations than GNU make, like BSD make or nmake used in Visual Studio ([according to Wikipedia](https://en.wikipedia.org/wiki/Make_(software))).

I tried searching for an alternative icon that carries the meaning of a task runner or something, but I couldn’t find one I’m satisfied with… (you can see them here: https://www.nerdfonts.com/cheat-sheet)

ariasuni:
What about `\u{f489}` (` `, `nf-oct-terminal`), since a `Makefile` is designed to run shell commands?

joshuagawley:
Apologies for the delay; I had changed my username in the meantime and so my commit history is somewhat unclean (the real change is in the second commit, where I changed the icon to the terminal icon).

ariasuni:
Could you rebase your work, or cherry-pick your commits on top of master? I’m not sure I want to merge those 18 commits. :sweat_smile:

ariasuni:
#945 was merged instead.

Freed-Wu:
> What about `\u{f489}` (` `, `nf-oct-terminal`), since a `Makefile` is designed to run shell commands?

How about  `\u{eb6d}`,  `\u{e20f}`,  `\u{f425}`? Makefile is not shell, Use same icon as shell is confusing. A icon containing tool may be good.

PR #924, opened 2021-08-06
Added -E --only-files filtering option 

not-a-duck:
-F was already taken, -E seemed like a good enough compromise.
There's a funny effect when both `--only-files`, `--only-dirs` are used.
For more clarity the options could be mutually exclusive, or this behaviour could simply be documented.


dcorbe:
Thanks for this.  I managed to delete a whole bunch of rigamarole from my zshrc by applying this PR to my fork.

PR #926, opened 2021-08-07
Update zsh completions for '--color'

Freed-Wu:
Now, when i press <kbd>Tab</kbd> after --color, it will
```zsh
❯ exa --color=always
(when)
always  auto    never
```

i think it will be better.


ariasuni:
Good catch, thanks!

PR #936, opened 2021-08-19
Make buildable on aarch64

sjames:
Hi,

I had to make this change to make it build on aarch64.

char is an u8 on aarch64. Use the low level char representation
instead of i8 to make it build on Arm64.

Regards,
Sojan


ariasuni:
I fixed it in a simpler way: https://github.com/ogham/exa/commit/8ad8b3357314ee743785902d8d9bc795c044ee29. Thanks anyway!

Issue #930, opened 2021-08-13
Add compression icon to .tXX files


izhakjakov:
- [ ] recognize `tgz` files as backup file similar to `tar.gz` files.
	- https://fileinfo.com/extension/tgz
	- https://www.file-extension.org/extensions/tgz
	- https://file.org/extension/tgz
	
- [ ] recognize `txz` files as backup file similar to `tar.xz` files.
	
	- https://fileinfo.com/extension/txz
	- https://www.file-extension.org/extensions/txz
	- https://file.org/extension/txz
	
- [ ] recognize `tlz` files as backup file similar to `tar.lz` files.
	- https://fileinfo.com/extension/tlz
	- https://www.file-extension.org/extensions/tlz
	- https://file.org/extension/tlz

izhakjakov:
Or anything else if I missed.
I noticed that `tbz` and `tbz2` are already recognized

jim4067:
working on this 👍🏾

izhakjakov:
thanks :)

izhakjakov:
Looks like it is no longer an issue
Thanks you jim4067 and ariasuni

PR #938, opened 2021-08-19
fix: add compression icon to .tXX files

jim4067:
#930 

Compression icons for .tXX files


src/output/icons.rs
```diff
@@ -302,6 +302,9 @@ pub fn icon_for_file(file: &File<'_>) -> char {
             "svg"           => '\u{f1c5}', // 
             "swift"         => '\u{e755}', // 
             "tar"           => '\u{f410}', // 
+            "tar.gz"        => '\u{f410}', // 
+            "tar.lz"        => '\u{f410}', // 
+            "tar.xz"        => '\u{f410}', // 
```
izhakjakov, line 307: I think "tgz", "tlz", "txz" should be added as well :)

ariasuni: They should added *instead*. Here, only the part after the last dot is matched against, and gz, lz and xz are already in the list.

src/info/filetype.rs
```diff
@@ -80,7 +80,8 @@ impl FileExtensions {
         file.extension_is_one_of( &[
             "zip", "tar", "Z", "z", "gz", "bz2", "a", "ar", "7z",
             "iso", "dmg", "tc", "rar", "par", "tgz", "xz", "txz",
-            "lz", "tlz", "lzma", "deb", "rpm", "zst",
+            "lz", "tlz", "lzma", "deb", "rpm", "zst", "tar.gz",
+            "tar.xz", "tar.lz",
```
ariasuni, line 84: The extension extracted by exa is only the part after the last dot, and gz, lz and xz are already in the list. So you should remove this change.

ariasuni (changes requested):

src/info/filetype.rs
jim4067: Noted

src/output/icons.rs
izhakjakov: Hi @ariasuni I think @jim4067 forgot to mention that he already addressed your comment in commit 437cd5b7f40ba1b704de211300b2eed42c93e71b

izhakjakov:
Are there any comments on changes made to commit 437cd5b7f40ba1b704de211300b2eed42c93e71b ?

codeswhite:
Missing `.lz4`

jim4067:
> Missing `.lz4`

I checked and  the last PR made was for frequently used file types. @ariasuni what do you think?

ariasuni:
oops I merged the other PR, but I’m pretty sure `.lz4` should be supported, especially since `.lz` is already supported. If you could create an new PR adding it in both places (`src/output/icons.rs` and `src/info/filetype.rs`), it would be perfect!

jim4067:
Okay. I'll make a PR for it too 👍🏾

PR #941, opened 2021-08-25
Show mounted filesystems

daviessm:
Introduce underlining of output for directories which have filesystems mounted on them. In long view the mounted file system information is displayed as `{/dev/mounted_device (filesystem)}`

This introduces the groundwork for identifying `btrfs` subvolumes which will be sent in a separate PR.

Happy to change the display of mount details on feedback.

Supersedes #452


daviessm:
Examples: First run shows a mounted `btrfs` filesystem with the root subvolume `/`. The second run shows three subvolumes contained within that subvolume.
Screenshot_20210825_162239

Issue #774, opened 2020-12-23
ln=target not working in LS_COLORS


acdifran:
version: 0.9.0
OS: macOS Big Sur 11.1

I'm using the LS_COLORS settings I've been using previously with "ls", which includes "ln=target" so that links are colored to match the target. However, this doesn't seem to work with exa.

You can see in the screenshot below that the color is correct for ls, but not for exa.

<img width="662" alt="Screen Shot 2020-12-23 at 12 18 10 AM" src="https://user-images.githubusercontent.com/25591626/102975437-eca57300-44b4-11eb-8ab6-badab78dbe11.png">

ariasuni:
I didn’t know it was possible to do this in GNU `ls` and it seems totally undocumented, but I can attest that’s indeed what `ls` does by looking at [its source code](https://github.com/coreutils/coreutils/blob/06f70d9ee69163fb2b18647963d6d6e81b6bd35d/src/ls.c#L2827). It is indeed not supported (yet).

PR #960, opened 2021-09-24
Support ln=target in LS_COLORS (#774)

LarsHaalck:
It is allowed to set `ln=target` instead of a color (e.g. [like this](https://github.com/trapd00r/LS_COLORS/blob/master/LS_COLORS#L42)), which was not supported by exa.

Setting `ln=target` highlights a link in the same color as the referred file. Dangling/orphaned links are always colored using the specified `or=` color.


jbrubake:
When I built `exa` using this patch, it does properly color symlinks, but for some reason certain combinations of terminal size (width I think, not height) and directory contents cause `exa` to always use the `--oneline` listing style.

I can't figure out what the exact situation is and looking at the patch, I can't see why it would have that effect, but it does @LarsHaalck

LarsHaalck:
You are right, but I think this is an issue with a previous commit and not with this PR. I also have this when building the exa from source without my PR.

jbrubake:
I didn't see how it could be related to your code. I should have checked to see if their were commits between the latest release and HEAD. I'll take a look later.

Nice work on the symlink patch. Since I found that feature of `ls` I find not having it annoying

PR #968, opened 2021-10-15
add icon for .bats files

docwhat:
This adds the same icon as BASH for `.bats` files.

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -118,6 +118,7 @@ pub fn icon_for_file(file: &File<'_>) -> char {
             "bash_profile"  => '\u{f489}', // 
             "bashrc"        => '\u{f489}', // 
             "bat"           => '\u{f17a}', // 
+            "bats"          => '\u{f489}', // 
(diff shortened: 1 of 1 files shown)


ariasuni:
I suppose you’re talking that? https://github.com/bats-core/bats-core

Well then you’ve just used the icon than `.bat`, which is to say the Windows icon, and it looks like it should use like you say, the same icon as Bash.

src/output/icons.rs
```diff
@@ -118,6 +118,7 @@ pub fn icon_for_file(file: &File<'_>) -> char {
             "bash_profile"  => '\u{f489}', // 
             "bashrc"        => '\u{f489}', // 
             "bat"           => '\u{f17a}', // 
+            "bats"          => '\u{f489}', // 
```
ariasuni, line 121: Ah I see, the comment is wrong but the code is correct.
```suggestion
            "bats"          => '\u{f489}', // 
```

ariasuni:
Thanks for your contribution!

PR #971, opened 2021-10-20
Add changelog as generated by `cargo changelog`

Byron:
`cargo smart-release` has a companion to help create changelogs from git-conventional messages in the git history, when used, and uses them to create a changelog-scaffolding which would then be filled in by hand. This goes as far as rejecting fully generated releaese notes when trying to release with `cargo smart-release`.

Both go hand-in-hand, and `smart-release` creates the same tag names as `cargo release` it's very similar, accept that `smart-release` will also write release notes (without links) to the tag annotation, and creates a github release as well if the `gh` tool is available.

[gitoxide already uses it](https://github.com/Byron/gitoxide/releases).

Please feel free to close as nobody asked for that, it's merely an attempt to share something I always wanted to have :D.


ariasuni:
I thought I had answered this… the generated changelog is not very useful, and for now the changelogs are generated by hand since we don’t use any sort of convention, so let’s close this for now (but maybe we’ll want something more automated in the future… thanks for sharing).

Issue #955, opened 2021-09-17
`$COLUMNS` will cause exa to colorize output when stdout is not connected with a tty


kidonng:
While using the workaround provided in https://github.com/ogham/exa/issues/879#issuecomment-845233451 to hide symlinks, I noticed that specifying `$COLUMNS` will cause exa to colorize output when stdout is not connected with a tty:

```sh
$ exa -a | cat -v
.config
.local

$ COLUMNS=1 exa -a | cat -v
^[[1;38;2;46;125;233m.config^[[0m
^[[1;38;2;46;125;233m.local^[[0m

# --color never still works though
$ COLUMNS=100 exa -a --color never  | cat -v
.config  .local
```

ariasuni:
Can confirm, using `COLUMNS` makes exa believes there’s a terminal at the other end. It should probably decorrelate «having a guessed/given number of columns» and «I am displaying to a terminal».

PR #975, opened 2021-10-31
Don't assume connection to a tty when number of columns is set

goth-turtle:
This fixes the colorization issue from #955. Im not sure about introducing an unsafe block for that libc call though, can I get an opinion on that?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -66,7 +66,7 @@ fn main() {
             let writer = io::stdout();
 
             let console_width = options.view.width.actual_terminal_width();
-            let theme = options.theme.to_theme(console_width.is_some());
+            let theme = options.theme.to_theme(unsafe {libc::isatty(1)} != 0);
             let exa = Exa { options, writer, input_paths, theme, console_width, git };
 
             match exa.run() {


ariasuni:
I think the `unsafe` block is fine, but I don’t think this call is portable e.g. on Windows which we want to support in the future.

I think that we should leverage the fact that the `terminal_size` is cross-platform and that the function `terminal_size()` returns `None` if there’s no TTY, and the call to `terminal_size()` in [`src/output/mod.rs`](https://github.com/ogham/exa/blob/b18e93d283d388ee42d8f0b73f36e546111324ce/src/output/mod.rs#L21) should be extracted and used in `main.rs`.

goth-turtle:
Im not sure if I understood you correctly, but for now I just replaced the `isatty()` call with `terminal_size()`.

I was also wondering if adding a testcase for this bug would be helpful? Ive looked into that already, but I dont think I entirely understand how to add a test yet, so I would need some help if you think I should add one.

ariasuni:
I suppose it’s something that you could test with Vagrant. See https://github.com/ogham/exa#testing-with-vagrant. Then it’s possible to add a test where exa is piped. Let me know if you want to do this or if you have questions.

I could do it but as you may have noticed by the time I took to comment your PR, I don’t have much time for exa these days, so it would be cool if you take a stab at it.

PR #979, opened 2021-11-24
Fix issue 887: can't compile on arm or aarch64

lefth:
Thanks to neepoo for starting this change. This PR builds on that, but is simpler and also supports arm.


ariasuni:
I fixed it in a simpler way: 8ad8b33. Thanks anyway!

PR #982, opened 2021-11-25
NO_COLOR environment variable support

### LINKED ISSUE #788: Add NO_COLOR support
Check the NO_COLOR environment variable to turn off color support.
See https://no-color.org.

paper-lark:
Fixes #788

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ parts
 prime
 stage
 *.snap
+
+# IntelliJ IDEA files
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,3 @@ parts
 prime
 stage
 *.snap
-
-# IntelliJ IDEA files
-.idea
--- a/man/exa.1.md
+++ b/man/exa.1.md
@@ -224,6 +224,12 @@ Specifies the number of spaces to print between an icon (see the ‘`--icons`’
 
 Different terminals display icons differently, as they usually take up more than one character width on screen, so there’s no “standard” number of spaces that exa can use to separate an icon from text. One space may place the icon too close to the text, and two spaces may place it too far away. So the choice is left up to the user to configure depending on their terminal emulator.
 
+## `NO_COLOR`
+
+Disables ANSI colour in the output (regardless of its value). Can be overridden by `--color` option.
+
+See `https://no-color.org/` for details.
+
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -19,3 +19,6 @@ parts
 prime
 stage
 *.snap
+
+# IntelliJ IDEA files
+.idea
```
ariasuni, line 24: I think you should put this kind of stuff in your global gitignore.

man/exa.1.md
```diff
@@ -224,6 +224,12 @@ Specifies the number of spaces to print between an icon (see the ‘`--icons`’
 
 Different terminals display icons differently, as they usually take up more than one character width on screen, so there’s no “standard” number of spaces that exa can use to separate an icon from text. One space may place the icon too close to the text, and two spaces may place it too far away. So the choice is left up to the user to configure depending on their terminal emulator.
 
+## `NO_COLOR`
+
+Disables ANSI colour in the output (regardless of its value). Can be overridden by `--color` option.
```
ariasuni, line 229: I’m not sure that specifying «ANSI» is useful. Also, I think adding a link to the website would be nice.

```suggestion
Disables colours in the output (regardless of its value). Can be overridden by `--color` option. See https://no-color.org/.
```

ariasuni (changes requested):
Nice, thanks! I’d like a very few things changes before I merge this, please look at my comments!

man/exa.1.md
paper-lark: Thanks, I removed ANSI word. As for the link it was already included on the next line.

paper-lark:
@ariasuni Thank you for the feedback! I've fixed everything based on your comments

ariasuni:
Thanks for you contribution :slightly_smiling_face:

ariasuni:
Would you like @paper-lark to contribute to the no-color website and add exa there as well?

paper-lark:
> Would you like @paper-lark to contribute to the no-color website and add exa there as well?

Sure. Though maybe we should release a new version of exa with NO_COLOR support first

ariasuni:
I’m not in charge of making new releases, and it can take a while, so I figured out it would be nice to add it. And when a new release of exa is published, it’s possible that we forget to update the no-color website, so…

paper-lark:
Oh, sure, I'll update the website then

PR #993, opened 2021-12-15
Adding icons for .part and .torrent files

cab-1729:
I set the icons similar to what I found in this XFCE theme : 

https://www.pling.com/s/XFCE/p/1463720

Which looks like : 

A similar change has been accepted here : 

https://github.com/alexanderjeurissen/ranger_devicons/pull/101


ariasuni:
Thanks!

PR #1006, opened 2022-01-08
Add git icon to .gitignore_global file

duboiss:
We already have it for `gitignore_global` but not `.gitignore_global`.

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -57,6 +57,7 @@ lazy_static! {
         m.insert(".gitconfig", '\u{f1d3}'); // 
         m.insert(".github", '\u{f408}'); // 
         m.insert(".gitignore", '\u{f1d3}'); // 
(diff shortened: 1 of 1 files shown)


ariasuni:
I’m not sure I understand why we support `gitignore_global` in the first place; maybe it makes on macOS or Windows?? I guess `.gitignore_global` makes sense, even though it’s not an hardcoded name in Git or even recommended in its documentation, because it’s almost standard now.

As for supporting, files starting with `.git`… well, I think people should just put their configuration in `~/.config/git/{config,ignore,etc.}` (at least on Linux, not sure how it work for other systems); I don’t think starting to make exceptions for Git is very useful here.

duboiss:
> I’m not sure I understand why we support gitignore_global in the first place; maybe it makes on macOS or Windows??
I also do not know

> because it’s almost standard now
Completely

duboiss:
Any update ?

PR #1011, opened 2022-01-14
Make it easy to copy and paste

WesleyBatista:
Github shows a copy button when mouse hovering the code blocks and the `$` signs makes it a bit inconvenient.


ariasuni:
Nice catch!

PR #1017, opened 2022-01-26
Change icon for gradle files to Java

pe:
Gradle files are for the whole Java ecosystem. Not just Android.


ariasuni:
I think you should use one of the Java icons, let’s say U+E256 since it’s the first one in code point order. Could you change the other icons using U+E204 to U+256 (if it renders)? That would be much appreciated.

pe:
Done. And it does render. The icon was already used for other java-related files :-)

pe:
Anything I can help with do to get this PR merged? Otherwise, I'll close it in about one week.

Issue #1014, opened 2022-01-25
On Ubuntu 20.10 (Groovy Gorilla) and later, package link is broken


Rassska:
Hi there!
Would be great to resolve that!

Thanks, Rasul!

SOSANA:
any updates? tried to down with Ubuntu and no success

ariasuni:
@SOSANA What are you talking about? Here we’re talking about the link in the README. What’s your version of Ubuntu and what steps did you follow?

SOSANA:
Please disregard as I am using ubuntu 20.04 and miss read both documentation and title of this issue.

manaslupisani:
via https://the.exa.website/#installation "On Ubuntu 20.10 (Groovy Gorilla) and later, install the [exa](https://packages.ubuntu.com/groovy/exa) package.", where https://packages.ubuntu.com/groovy/exa results in a 404. 

So currently, exa packages in Ubuntu for the last 5 versions: 
[Doesn't exist] https://packages.ubuntu.com/xenial/exa
[Doesn't exist] https://packages.ubuntu.com/bionic/exa
[Doesn't exist] https://packages.ubuntu.com/focal/exa
[Exists!] https://packages.ubuntu.com/impish/exa
[Exists!] https://packages.ubuntu.com/jammy/exa

How about for any unsupported version, either a) instruct how to download and install the .deb or b) point to the guide on installing from src?

ariasuni:
Yeah the website is unmaintained, but the README link will be correct after I merge #1020.

ariasuni:
Closed in favor of https://github.com/ogham/exa/issues/757

exa is unmaintained, use [the active fork eza](https://github.com/eza-community/eza) instead (see https://github.com/ogham/exa/issues/1243).

PR #1020, opened 2022-02-02
Fix broken link to Ubuntu Package

reishoku:
This PR includes:
- Fix broken link to Ubuntu Package

It seems that Ubuntu provides `exa` package only for Ubuntu 21.04 and 21.10 and 22.04 (hirsute, impish, jammy). For reference, see https://packages.ubuntu.com/search?keywords=exa&searchon=names&exact=1&suite=all&section=all .

closes: https://github.com/ogham/exa/issues/1014

-------------------------

Signed-off-by: KOSHIKAWA Kenichi <reishoku.misc ~~@~~ pm.me>

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ On openSUSE, install the [`exa`](https://software.opensuse.org/package/exa) pack
 
 ### Ubuntu
 
-On Ubuntu 20.10 (Groovy Gorilla) and later, install the [`exa`](https://packages.ubuntu.com/groovy/exa) package.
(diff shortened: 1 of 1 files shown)


ariasuni:
I don’t think you actually need to change the text, just the link. Also, maybe use https://packages.ubuntu.com/jammy/exa instead, because it’s the link for the next LTS (that’ll be released very soon) so it’ll stay valid for a longer time.

ariasuni:
I did the change in https://github.com/ogham/exa/commit/dceca33779e92f91589027650f9938d5360e16ba. Thanks anyway!

PR #1046, opened 2022-03-18
Add JPEG 2000 filetype

sorairolake:


ariasuni:
According to Wikipedia:

> The standardized [filename extension](https://en.wikipedia.org/wiki/Filename_extension) is **.jp2** for [ISO](https://en.wikipedia.org/wiki/International_Organization_for_Standardization)/[IEC](https://en.wikipedia.org/wiki/International_Electrotechnical_Commission) 15444-1 conforming files and **.jpx** for the extended part-2 specifications, published as ISO/IEC 15444-2.

So I think it would be relevant to include **.jpx** as well.

sorairolake:
@ariasuni Added `.jpx`.

PR #1049, opened 2022-03-29
Adds "almost all" flag (-A) for compatibility with zsh's la alias

ptcar2009:


ariasuni (changes requested):
Well, the problem here is that your code change doesn’t handle the conflict between `-a` and `-A`. In exa in general, when using two contradictory flags, the latter is used. But here, `exa -aaA` would mean `exa -aa`, which is not quite what we want.

I think you should look at how `--oneline` overrides `--long`, for example.

Also, `-AA` should produce an error, see:
https://github.com/ogham/exa/pull/1049/files#diff-e691d70a9f56ceed21c51d53fe510b25dcc42814c641795e327ff141aaf704bdR172-R174

ariasuni:
Closing this since exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and this has been done in [the active fork eza](https://github.com/eza-community/eza).

(see https://github.com/eza-community/eza/pull/43)

PR #1050, opened 2022-04-01
Fix `just` command's unavailable option

ewrenge:
`just --tasks` option is seems to not defined
So, I think that's probably meaning [`just --list`](https://github.com/casey/just#listing-available-recipes)


ariasuni:
Yep, not sure why it was `--tasks` because I can’t find any ref, it should have been changed a long time ago. Now that Just is stable, it shouldn’t change again. Thanks!

PR #1064, opened 2022-05-10
Add `-A` option for more compatibility with `ls`

### LINKED ISSUE #497: No -A switch
I like exa, but I use it with a ` alias ls='exa -Bg'`, to act more like ls.

The problem is that I have a decade of muscle memory typing `ls -lA`.

I know that exa's `-la` switch works in the same way as `ls -lA`, as it doesn't include the current and parent dot directories, but I think it would be cool if the -A switch was included as an alias for -a, just for compatibility and people like me.

kejadlen:
Fixes #497.

I arbitrarily named the option `--ls-all` since naming is hard and I couldn't think of something better.

### DIFF
--- a/src/options/filter.rs
+++ b/src/options/filter.rs
@@ -149,27 +149,35 @@ impl DotFilter {
 
     /// Determines the dot filter based on how many `--all` options were
     /// given: one will show dotfiles, but two will show `.` and `..` too.
+    /// --ls-all is equivalent to --all, included for compatibility with
+    /// `ls -A`.
     ///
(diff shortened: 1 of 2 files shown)


Bond-009:
It's called `--almost-all` in GNU ls

ariasuni:
Closing this since exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and this has been done in [the active fork eza](https://github.com/eza-community/eza).

(see https://github.com/eza-community/eza/pull/43)

PR #1076, opened 2022-06-17
Added icons for .pm similar to .pl

cab-1729:
Perl modules end with the _.pm_ extension. These are files imported by Perl scripts as objects. Most file managers I know display the same icon for .pm and .pl files.

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -272,8 +272,10 @@ pub fn icon_for_file(file: &File<'_>) -> char {
             "pdf"           => '\u{f1c1}', // 
             "php"           => '\u{e73d}', // 
             "pl"            => '\u{e769}', // 
+            "plx"           => '\u{e769}', // 
             "pm"            => '\u{e769}', // 
             "png"           => '\u{f1c5}', // 
+            "pod"           => '\u{e769}', // 
             "ppt"           => '\u{f1c4}', // 
             "pptx"          => '\u{f1c4}', // 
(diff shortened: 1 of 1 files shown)


ariasuni:
Hi, there is already a similar pull request open but its author seems not responsive.

Do you think you could take the [types mentioned here](https://github.com/ogham/exa/pull/880#issuecomment-869178331) into account, and add the relevant ones to your pull request?

cab-1729:
I've added the same icon for ```pm```, ```t```, ```pod```, and ```plx```.

I think we should not concern ourselves with the fact that ```plx``` is used in legacy code, given that a large portion of programmers interacting with perl are dealing with legacy code.

```t``` is just perl code used for testing.

You are right about ```pod``` being documentation, but _Plain Old Documentation_(syntax used by ```pod``` files) is a sort of markup language often used inside of perl scripts instead of separate files. The perl interpreter ignores what is written with this syntax (similar to comments), hence ```.pod``` files are technically valid perl code. So it might make sense for them to have the same icon. Besides that, there isn't really another good alternative I can think of.

With regard to ```.xs``` files. These files allow perl to interact with C/C++ subroutines. It isn't really perl code, but the syntax is sort of like a combination of C and Perl. So I didn't give it the same icon. I think it is better to ignore ```xs``` files for now, because I couldn't find an icon that seems appropriate for those files.

ariasuni:
OK, thanks a lot!

PR #1085, opened 2022-07-10
add Zig module icons

Syuq:


cab-1729:
[Nerd fonts](https://github.com/ryanoasis/nerd-fonts/tree/gh-pages) as of now does not support an icon for ```Zig```. I think it would be better to first make a pull request on their side and add such an icon prior to adding it to this project.
_Unicode 21af_ is not a part of _Nerd Fonts_. [According to a maintainer, this project likely does not use icons that do not come from _Nerd Fonts_.](https://github.com/ogham/exa/pull/1057#issuecomment-1113515790).

On top of that, your commit seems to add the icon for a ```.zig``` hidden file instead of adding the icon to a file with the extension ```.zig```.

Issue #519, opened 2019-07-10
-t


heptal:
This -t thing really is a problem, it's flow-breaking and existing `ls`-based alias-breaking. It's basically infuriating because everything else works so well. Even an env variable I could set would be great.

```
Flag -t needs a value (choices: modified, accessed, created)
To sort newest files last, try "--sort newest", or just "-snew"
```

https://github.com/ogham/exa/issues/294#issuecomment-453884188
https://github.com/ogham/exa/issues/284#issuecomment-329536899

varenc:
agreed 100% on this.  I love exa! But the lack of `-t` really breaks my flow.  Whenever this happens, I find myself consistently using `\ls -ltr` to get around my `ls=exa` alias.  The `-snew` option is nice, but longer to type and I find multiple character single dash options a bit confusing.   (it looks like each character is its own seperate option). The helpful error message also makes it clear exa knows what I'm trying to do.

For now, I've set up a little function that `ls` maps to.  Usually, this function just runs `exa` directly, but it also checks if the args contain a single dash `t` option and if they do it removes `t` and adds `-snew` to the args passed to exa.  It's quite hacky and bugs out sometimes, but it is generally what I want.  I like [this comment](https://github.com/ogham/exa/issues/284#issuecomment-329530883) which suggests a `--legacy` option to enable more `ls` compliant behavior like this.

RandomErrorMessage:
Check out LSD for a similar tool that is actually a drop-in replacement for LS.

rwilhelm:
> Check out LSD for a similar tool that is actually a drop-in replacement for LS.

+1, LSD is really great. Highly recommended.

madebyollin:
Since I prefer `exa` to `lsd` apart from this particular bit of broken muscle memory...
```bash
ls() {
    if [ "$1" = "-ltr" ]; then
        exa -lsnew "${@:2}"
    elif [ "$1" = "-lrt" ]; then
        exa -lsnew "${@:2}"
    else
        exa "$@"
    fi
}
```
(In `.profile`, `.bashrc`, whatever, instead of the usual `alias ls=exa`)

kaushalyap:
@ogham Any plans to fix this incompatibility?

ariasuni:
[...]

PR #1102, opened 2022-08-09
Make `-t` work like ls when used with no args

WhyNotHugo:
Currently `exa` can't work as a drop-in `ls` replacement due to `-t`
having a very different behaviour.

However `-t` currently only work when used with a value, and is invalid
when used without any value.

This patch changes the behaviour of `-t` when used with no arguments to
behave exactly like it does for `ls`. When used with a value, it
continues behaving as in previous versions of `exa`.

Fixes: https://github.com/ogham/exa/issues/519


WhyNotHugo:
This isn't in a ready-to-merge state:

- Formatting needs to be fixed, but running `rust fmt` will mess up all the existing formatting which I'm sure would be unacceptable.
- There are no tests. After reading all the work put into test, I'm pretty sure that's mandatory for merging.

However, I'm presenting this here early mostly to hear feedback on the approach. I can't decide if this is a beautiful fix, a horrible hack, or somehow, both at the same time.

WhyNotHugo:
Oh, I'm not sure what should be done when `--sort` is passed **and** `-t` with no value is passed; it seems one would be redundant.

WhyNotHugo:
@ariasuni Any feedback for me here?

PR #1125, opened 2022-10-01
Upgrade both Rust version and edition (and fix some lints)

Tyrubias:
This PR updates `exa` to use Rust 1.63.0 as well as Rust Edition 2021. It also fixes a number of Clippy warnings. Note that we do not upgrade to Rust 1.64.0 here because it increases the `glibc` and Linux kernel requirements, which would break `exa` for users on older platforms without necessarily providing a benefit in terms of developer experience or new APIs.

This is my first Rust contribution. I did my best to follow [the atomic commits principle](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/) to make it easier to review my PR. I'm inexperienced with open-source contributions and would love any feedback you may have. Thank you!


src/options/filter.rs
```diff
@@ -294,7 +294,7 @@ mod test {
     mod ignore_patterns {
         use super::*;
         use std::iter::FromIterator;
-        use glob;
+
```
ariasuni, line 297: You can remove the line, there should only be one line between the `use` block and the next block.

src/options/parser.rs
```diff
@@ -91,7 +91,7 @@ pub enum Strictness {
 
 /// Whether a flag takes a value. This is applicable to both long and short
 /// arguments.
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
```
ariasuni, line 94: Maybe change this to:

```suggestion
#[derive(PartialEq, Eq, Debug, Copy, Clone)]
```

so that everything use the same order for derive.

ariasuni (approved):
Thanks a lot for trying to improve exa!

You will need to rebase your changes, because there were a few changes since you made this pull request. If you make the necessary changes, I can merge it immediately.

ariasuni:
Could you update `.github/workflows/unit-tests.yml` for CI to work as well? (editing the 1.56.1 → 1.63.0)

PR #1141, opened 2022-10-30
git: Use GIT_DIR env var to find the repo

### LINKED ISSUE #722: FR: Respect environment variables GIT_DIR and GIT_WORK_TREE
Thanks for creating such an awesome upgrade to the ls command!

I use git-bare to manage my dotfiles:
https://www.atlassian.com/git/tutorials/dotfiles
TL;DR we change git-dir and git-work-tree so $HOME acts as repo root in a more elegant way.

I'd like to get exa working with this setup for its `--git` flag (using a function wrapper). Would it be possible to let exa read `GIT_DIR` and `GIT_WORK_TREE` and pass them to git when they exists and where appropriate?

iamkroot:
`GIT_DIR` overrides everything else.
So even if there is a valid git repo in the current dir, git would use the repo at `GIT_DIR`.

Closes #722

### DIFF
--- a/src/fs/feature/git.rs
+++ b/src/fs/feature/git.rs
@@ -166,7 +166,16 @@ impl GitRepo {
             Ok(r) => r,
             Err(e) => {
                 error!("Error discovering Git repositories: {:?}", e);
-                return Err(path);
+                match git2::Repository::open_from_env() {
+                    Ok(r) => r,
+                    Err(e) => {
+                        // anything other than NotFound implies GIT_DIR was set and we got actual error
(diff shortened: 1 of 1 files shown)


iamkroot:
@ogham Could you please take a look at this if you get time? It is a pretty small change

ariasuni:
Closing this, since exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and this has been done in [the active fork eza](https://github.com/eza-community/eza). Thanks!

Issue #1145, opened 2022-12-10
Makefiles spelled with a lowercase "m" do not receive a special color


saltedcoffii:
Makefile spelled with a lowercase "m" don't get a special highlight and color like makefiles with a capital "M". See screenshots below, with problem encircled. The tarball in the circle is the source code of unrar 6.2.2, found [here](https://www.rarlab.com/rar/unrarsrc-6.2.2.tar.gz). I wasn't sure if it was just GNU make that responds to uncapitalized makefiles, however apparently all of GNU make, bmake, and kati work for these. I'm not sure if they're standard.

#### Expected behavior (for comparison)
![exa-Makefile](https://user-images.githubusercontent.com/55339220/206852441-19dd746c-d1ef-4e1a-97b8-274e276ad4fc.png)

#### Problematic behavior
![exa-makefile](https://user-images.githubusercontent.com/55339220/206852440-c589eeb0-702a-4f80-b8ea-db483cfbbc90.png)

```
$ exa --version
exa - list files on the command-line
v0.10.1 [+git]
https://the.exa.website/
```
Operating system: Arch Linux rolling
Architecture: x86_64

ariasuni:
exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and this has been fixed in [the active fork eza](https://github.com/eza-community/eza).

PR #1146, opened 2022-12-14
fix issue 1145

tommady:
fix issue [1145](https://github.com/ogham/exa/issues/1145)

according to the GNU [doc](https://web.mit.edu/gnu/doc/html/make_3.html#:~:text=By%20default%2C%20when%20make%20looks,makefile'%20or%20%60Makefile'.)

```
The default makefile names `GNUmakefile', `makefile' and `Makefile'
```

results
image
image

### DIFF
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -23,14 +23,42 @@ impl FileExtensions {
     fn is_immediate(&self, file: &File<'_>) -> bool {
         file.name.to_lowercase().starts_with("readme") ||
         file.name.ends_with(".ninja") ||
-        file.name_is_one_of( &[
-            "GNUmakefile", "Makefile", "makefile", "Cargo.toml", "SConstruct", "CMakeLists.txt",
-            "build.gradle", "pom.xml", "Rakefile", "package.json", "Gruntfile.js",
-            "Gruntfile.coffee", "BUILD", "BUILD.bazel", "WORKSPACE", "build.xml", "Podfile",
-            "webpack.config.js", "meson.build", "composer.json", "RoboFile.php", "PKGBUILD",
-            "Justfile", "Procfile", "Dockerfile", "Containerfile", "Vagrantfile", "Brewfile",
-            "Gemfile", "Pipfile", "build.sbt", "mix.exs", "bsconfig.json", "tsconfig.json",
-        ])
+        [
+            "BUILD",
+            "BUILD.bazel",
+            "Brewfile",
+            "CMakeLists.txt",
+            "Cargo.toml",
+            "Containerfile",
+            "Dockerfile",
+            "GNUmakefile",
+            "Gemfile",
+            "Gruntfile.coffee",
+            "Gruntfile.js",
+            "Justfile",
+            "Makefile",
+            "PKGBUILD",
+            "Pipfile",
+            "Podfile",
+            "Procfile",
+            "Rakefile",
+            "RoboFile.php",
+            "SConstruct",
+            "Vagrantfile",
+            "WORKSPACE",
+            "bsconfig.json",
+            "build.gradle",
+            "build.sbt",
+            "build.xml",
+            "composer.json",
+            "makefile",
+            "meson.build",
+            "mix.exs",
+            "package.json",
+            "pom.xml",
+            "tsconfig.json",
+            "webpack.config.js",
+        ].binary_search(&file.name.as_str()).is_ok()
     }
 
     fn is_image(&self, file: &File<'_>) -> bool {
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -22,10 +22,9 @@ impl FileExtensions {
     #[allow(clippy::case_sensitive_file_extension_comparisons)]
     fn is_immediate(&self, file: &File<'_>) -> bool {
         file.name.to_lowercase().starts_with("readme") ||
-        file.name.to_lowercase().starts_with("makefile") ||
         file.name.ends_with(".ninja") ||
         file.name_is_one_of( &[
-            "GNUmakefile", "Cargo.toml", "SConstruct", "CMakeLists.txt",
+            "GNUmakefile", "Makefile", "makefile", "Cargo.toml", "SConstruct", "CMakeLists.txt",
             "build.gradle", "pom.xml", "Rakefile", "package.json", "Gruntfile.js",
             "Gruntfile.coffee", "BUILD", "BUILD.bazel", "WORKSPACE", "build.xml", "Podfile",
             "webpack.config.js", "meson.build", "composer.json", "RoboFile.php", "PKGBUILD",
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -21,44 +21,45 @@ impl FileExtensions {
     /// in directories full of source code.
     #[allow(clippy::case_sensitive_file_extension_comparisons)]
     fn is_immediate(&self, file: &File<'_>) -> bool {
-        file.name.to_lowercase().starts_with("readme") ||
-        file.name.ends_with(".ninja") ||
-        [
-            "BUILD",
-            "BUILD.bazel",
-            "Brewfile",
-            "CMakeLists.txt",
-            "Cargo.toml",
-            "Containerfile",
-            "Dockerfile",
-            "GNUmakefile",
-            "Gemfile",
-            "Gruntfile.coffee",
-            "Gruntfile.js",
-            "Justfile",
-            "Makefile",
-            "PKGBUILD",
-            "Pipfile",
-            "Podfile",
-            "Procfile",
-            "Rakefile",
-            "RoboFile.php",
-            "SConstruct",
-            "Vagrantfile",
-            "WORKSPACE",
-            "bsconfig.json",
-            "build.gradle",
-            "build.sbt",
-            "build.xml",
-            "composer.json",
-            "makefile",
-            "meson.build",
-            "mix.exs",
-            "package.json",
-            "pom.xml",
-            "tsconfig.json",
-            "webpack.config.js",
-        ].binary_search(&file.name.as_str()).is_ok()
+	file.name.to_lowercase().starts_with("readme")
+            || file.name.ends_with(".ninja")
+            || matches!(
+                file.name.as_str(),
+                "BUILD"
+                    | "BUILD.bazel"
+                    | "Brewfile"
+                    | "CMakeLists.txt"
+                    | "Cargo.toml"
+                    | "Containerfile"
+                    | "Dockerfile"
+                    | "GNUmakefile"
+                    | "Gemfile"
+                    | "Gruntfile.coffee"
+                    | "Gruntfile.js"
+                    | "Justfile"
+                    | "Makefile"
+                    | "PKGBUILD"
+                    | "Pipfile"
+                    | "Podfile"
+                    | "Procfile"
+                    | "Rakefile"
+                    | "RoboFile.php"
+                    | "SConstruct"
+                    | "Vagrantfile"
+                    | "WORKSPACE"
+                    | "bsconfig.json"
+                    | "build.gradle"
+                    | "build.sbt"
+                    | "build.xml"
+                    | "composer.json"
+                    | "makefile"
+                    | "meson.build"
+                    | "mix.exs"
+                    | "package.json"
+                    | "pom.xml"
+                    | "tsconfig.json"
+                    | "webpack.config.js"
+            )
     }
 
     fn is_image(&self, file: &File<'_>) -> bool {
--- a/src/info/filetype.rs
+++ b/src/info/filetype.rs
@@ -22,9 +22,10 @@ impl FileExtensions {
     #[allow(clippy::case_sensitive_file_extension_comparisons)]
     fn is_immediate(&self, file: &File<'_>) -> bool {
         file.name.to_lowercase().starts_with("readme") ||
+        file.name.to_lowercase().starts_with("makefile") ||
         file.name.ends_with(".ninja") ||
         file.name_is_one_of( &[
-            "Makefile", "Cargo.toml", "SConstruct", "CMakeLists.txt",
+            "GNUmakefile", "Cargo.toml", "SConstruct", "CMakeLists.txt",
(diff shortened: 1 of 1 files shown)


DriesMeerman (approved):
Might be nice to put the starts_with calls into an array of files with certain prefixes

src/info/filetype.rs
```diff
@@ -22,9 +22,10 @@ impl FileExtensions {
     #[allow(clippy::case_sensitive_file_extension_comparisons)]
     fn is_immediate(&self, file: &File<'_>) -> bool {
         file.name.to_lowercase().starts_with("readme") ||
+        file.name.to_lowercase().starts_with("makefile") ||
```
marbx, line 25: Hi @tommady,
wouldn't that also match files like `makefile.txt` or `makefile123`, which are no Makefiles?

tommady: I just followed the GNU doc default description as I mentioned in the PR.
Other requests I think is beyond this PRs scope.

marbx:
Could you please create a file makefile123 and run this PR, as shown in
your screenshots?
Because makefile123 startswith makefile the PR treats it as a makefile.

tommady:
> 

oh! I see what you meant ahah
sure will fix that, thank you!

tommady:
new result
image

marbx:
Your commit 92f8eba fixes what I meant.
Thank you for you swift reply :-)


I have a question regarding performance.

The function `name_is_one_of()` is called with 35 file names and is implemented with
https://github.com/ogham/exa/blob/f3ca1fe6f7d8242cee3d9fd602c96c9db041a7a0/src/fs/file.rs#L502


`choices` is the only parameter of the function and hold the file names:
https://github.com/ogham/exa/blob/f3ca1fe6f7d8242cee3d9fd602c96c9db041a7a0/src/fs/file.rs#L501


I am beginner of Rust, so reading the [Standard library](https://doc.rust-lang.org/std/?search=contains#containers-and-collections), `&[T]` is a  "shared slice".


The [method `contains()` of the slice primitive](https://doc.rust-lang.org/std/primitive.slice.html#method.contains) has a comment that says

> This operation is O(n).
> Note that if you have a sorted slice, binary_search may be faster.

Is this the method used?

"O(n)" means linear search. 
I have the impression that 35 file names could be large enough to justify sorting and using a binary search.

And for the sake of performance, sorting could be done by sorting the array in the soource.

What do you think?

tommady:
> Your commit [92f8eba](https://github.com/ogham/exa/commit/92f8eba90976d458e8dbfe51099bc689781468ce) fixes what I meant. Thank you for you swift reply :-)
> 
> I have a question regarding performance.
> 
> The function `name_is_one_of()` is called with 35 file names and is implemented with
> 
> https://github.com/ogham/exa/blob/f3ca1fe6f7d8242cee3d9fd602c96c9db041a7a0/src/fs/file.rs#L502
> 
> `choices` is the only parameter of the function and hold the file names:
> 
> https://github.com/ogham/exa/blob/f3ca1fe6f7d8242cee3d9fd602c96c9db041a7a0/src/fs/file.rs#L501
> 
> I am beginner of Rust, so reading the [Standard library](https://doc.rust-lang.org/std/?search=contains#containers-and-collections), `&[T]` is a "shared slice".
> 
> The [method `contains()` of the slice primitive](https://doc.rust-lang.org/std/primitive.slice.html#method.contains) has a comment that says
> 
> > This operation is O(n).
> > Note that if you have a sorted slice, binary_search may be faster.
> 
> Is this the method used?
> 
> "O(n)" means linear search. I have the impression that 35 file names could be large enough to justify sorting and using a binary search.
> 
> And for the sake of performance, sorting could be done by sorting the array in the source.
> 
> What do you think?

ya, I agree with your point.
I will do that!

BTW, I am a rust beginner too~
thanks for your suggestion!

marbx:
@tommady, I appreciate the collaboration!
The sorted, vertical list from ba9c762  is much more readable then the unsorted list across 6 lines, I think.

I also think that is goes without saying, that any new filename needs to be added at its sorted positions.
I leave it to the exa maintainers whether a comment like this would make sense: 
> // `binary_search()` requires sorted file names

marbx:
On a second thought:
the function `name_is_one_of()` is currently used only here and this PR would makes the function unused.
Unused code is a bad thing.


As there could be other users, I would like to suggest to move the binary search into the function, together with a break-even calculation.

Along the lines:
`if length of choices is below 20: contains(), else: binary_search()`

So the potential other users would profit, too.

Final argument: the  `name_is_one_of()` function is currently "too simple". It cyclomatic complexity is probably 1.
In other words: it could use some "flesh" :-)

marbx:
Also, we must not `use binary_search()` on only 35 file names!
It is 16% slower than `contains()`


I benchmarked `contains()` and `binary_search()` on 35 and 85 words:
https://github.com/marbx/LearnRust/blob/main/BenchmarkContainsBinarySearch/src/lib.rs


```
test tests::bench_binary_search_with_35_words ... bench:          25 ns/iter (+/- 0)
test tests::bench_binary_search_with_85_words ... bench:          33 ns/iter (+/- 1)
test tests::bench_contains_with_35_word       ... bench:          21 ns/iter (+/- 0)
test tests::bench_contains_with_85_words      ... bench:          58 ns/iter (+/- 5)
test tests::bench_xor                         ... bench:          93 ns/iter (+/- 2)
```


Break-even must be between 35 and 85, I will try to find it in the next days.

marbx:
Renaming the functions changes the benchmarked times!
```
test tests::bench_35_words_binary_search ... bench:          25 ns/iter (+/- 0)
test tests::bench_35_words_contains      ... bench:          25 ns/iter (+/- 0)
test tests::bench_85_words_binary_search ... bench:          33 ns/iter (+/- 1)
test tests::bench_85_words_contains      ... bench:          63 ns/iter (+/- 1)
test tests::bench_xor                    ... bench:          92 ns/iter (+/- 3)
```

tommady:
wow
so based on your comments, this PR's filenames are only 34 which means we should use the *contains* method right?

and thank you for your survey which made me learn so much.

marbx:
I thought so 1 hour ago, but I realized that in the "real world" there are much more files which are **not** in  the list. The number above come from a benchmark "1 file absent, 1 file present". 
This is unrealistic

So I changed the benchmark to: "4 files absent, 1 file present" and got:
```
test tests::bench_35_words_binary_search ... bench:          72 ns/iter (+/- 1)
test tests::bench_35_words_contains      ... bench:          78 ns/iter (+/- 8)
test tests::bench_85_words_binary_search ... bench:          99 ns/iter (+/- 2)
test tests::bench_85_words_contains      ... bench:         138 ns/iter (+/- 1)
```
This I understand, because  `contains()` must go though the whole list, before it can conclude that the file name is **absent**, so this benchmark hurts it a lot more than `binary_search()`.

As you see, **this** benchmark is in favor of `binary_search()` already for 35 file names.
I would say in the "real world", there are more then 4 times files outside of the list.

In short:  `binary_search()` is safer. 

Could you please run and play with the benchmark? I don't know how trustworthy the measurements are.

I just realize: for a directory (or tree listing) larger then 35, one would have to **reverse** the search and (binary) search 35 file names in the directory/listing.

If you have a directory of 100 files, exa currently  searches  100 times over a list of 35 entries.
While it is  enough to search 35 times in a list of 100 entries.

tommady:
or just a stupid idea, we can make the immediate files into a static hashmap which will lead every search O(1).

WDYT?

marbx:
Not stupid idea at all - I got the same.

A hashmap is the fastest of the three methods, even if created within each benchmark function. So hashmap is the way forward.
```
test tests::bench_35_words_binary_search         ... bench:          72 ns/iter (+/- 5)
test tests::bench_35_words_contains              ... bench:          69 ns/iter (+/- 27)
test tests::bench_35_words_hashmap_local_mutable ... bench:          57 ns/iter (+/- 1)
test tests::bench_85_words_binary_search         ... bench:          99 ns/iter (+/- 2)
test tests::bench_85_words_contains              ... bench:         138 ns/iter (+/- 2)
test tests::bench_85_words_hashmap_local_mutable ... bench:          57 ns/iter (+/- 4)
test tests::bench_xor                            ... bench:          93 ns/iter (+/- 3)
```
The measurement supports the O(1) expectation.

it is easy to setup a local, mutable hashmap.
I have not yet figured out how to setup a **static, global, unmutable** hashmap.

tommady:
I saw the exa already uses the dependency "lazy_static",
so ideally it is not hard to archive the goal, let me do a new commit!

marbx:
[Rust-PHF is a library to generate efficient lookup tables at compile time using perfect hash functions](https://docs.rs/phf/0.11.1/phf/).

Unfortunately,  PHF is  slower than the dynamically created hashmap!
```
test tests::bench_35_words_binary_search         ... bench:          74 ns/iter (+/- 2)
test tests::bench_35_words_contains              ... bench:          76 ns/iter (+/- 1)
test tests::bench_35_words_hashmap_local_mutable ... bench:          59 ns/iter (+/- 1)
test tests::bench_35_words_hashmap_static_phf    ... bench:          71 ns/iter (+/- 3)
test tests::bench_35_words_static_phf_set        ... bench:          70 ns/iter (+/- 1)
test tests::bench_85_words_binary_search         ... bench:          98 ns/iter (+/- 2)
test tests::bench_85_words_contains              ... bench:         228 ns/iter (+/- 24)
test tests::bench_85_words_hashmap_local_mutable ... bench:          59 ns/iter (+/- 1)
test tests::bench_85_words_hashmap_static_phf    ... bench:          71 ns/iter (+/- 1)
test tests::bench_85_words_static_phf_set        ... bench:          70 ns/iter (+/- 2)
test tests::bench_xor                            ... bench:          91 ns/iter (+/- 0)
```
What am I doing wrong?
https://github.com/marbx/LearnRust/blob/b41ff2cd5c4c73118c2fea9e2e8ba39f5e1a3425/BenchmarkContainsBinarySearch/src/lib.rs#L510

tommady:
or a more simple one?
try the simple match method.
https://www.reddit.com/r/rust/comments/5mnj3y/which_has_better_performance_a_hashmap_or_a/

marbx:
The first time, the lazy_static HashMap is as fast as the dynamically created one.
This makes lazy_static HashMap  the (current) winner.

Although, the second time there is no speed up.
This may have to to with the benchmark.

I'll try match


```
test tests::bench_35_words_binary_search         ... bench:          73 ns/iter (+/- 3)
test tests::bench_35_words_contains              ... bench:          94 ns/iter (+/- 10)
test tests::bench_35_words_hashmap_local_mutable ... bench:          60 ns/iter (+/- 3)
test tests::bench_35_words_hashmap_static_phf    ... bench:          70 ns/iter (+/- 2)
test tests::bench_35_words_lazy_hashset_1        ... bench:          60 ns/iter (+/- 4)
test tests::bench_35_words_lazy_hashset_2        ... bench:          60 ns/iter (+/- 1)
test tests::bench_35_words_static_phf_set        ... bench:          70 ns/iter (+/- 3)
test tests::bench_85_words_binary_search         ... bench:          99 ns/iter (+/- 3)
test tests::bench_85_words_contains              ... bench:         138 ns/iter (+/- 4)
test tests::bench_85_words_hashmap_local_mutable ... bench:          58 ns/iter (+/- 3)
test tests::bench_85_words_hashmap_static_phf    ... bench:          70 ns/iter (+/- 1)
test tests::bench_85_words_lazy_hashset_1        ... bench:          59 ns/iter (+/- 2)
test tests::bench_85_words_lazy_hashset_2        ... bench:          59 ns/iter (+/- 3)
test tests::bench_85_words_static_phf_set        ... bench:          70 ns/iter (+/- 2)
test tests::bench_xor                            ... bench:          91 ns/iter (+/- 0)


```

marbx:
Match  takes 0ns for 35 words, and 121ns for  85 words, double as for hash
This makes no sense.

```
test tests::bench_35_words_binary_search         ... bench:          72 ns/iter (+/- 3)
test tests::bench_35_words_contains              ... bench:          78 ns/iter (+/- 3)
test tests::bench_35_words_hashmap_local_mutable ... bench:          59 ns/iter (+/- 1)
test tests::bench_35_words_hashmap_static_phf    ... bench:          70 ns/iter (+/- 2)
test tests::bench_35_words_lazy_hashset_1        ... bench:          58 ns/iter (+/- 1)
test tests::bench_35_words_lazy_hashset_2        ... bench:          59 ns/iter (+/- 4)
test tests::bench_35_words_match                 ... bench:           0 ns/iter (+/- 0)
test tests::bench_35_words_static_phf_set        ... bench:          70 ns/iter (+/- 2)
test tests::bench_85_words_binary_search         ... bench:          99 ns/iter (+/- 4)
test tests::bench_85_words_contains              ... bench:         229 ns/iter (+/- 13)
test tests::bench_85_words_hashmap_local_mutable ... bench:          58 ns/iter (+/- 4)
test tests::bench_85_words_hashmap_static_phf    ... bench:          70 ns/iter (+/- 2)
test tests::bench_85_words_lazy_hashset_1        ... bench:          58 ns/iter (+/- 1)
test tests::bench_85_words_lazy_hashset_2        ... bench:          58 ns/iter (+/- 2)
test tests::bench_85_words_match                 ... bench:         121 ns/iter (+/- 6)
test tests::bench_85_words_static_phf_set        ... bench:          69 ns/iter (+/- 2)
test tests::bench_xor                            ... bench:          92 ns/iter (+/- 5)

```

marbx (commented):
matches! macro increases readability.

tommady:
Ya I think we should leave all those surveys for the maintainer to do the judgement.

Clear winner for this PR (which the file names are below 35) is the fastest one with no doubts.

ariasuni:
Does it do a difference in term of final executable size?

tommady:
> Does it do a difference in term of final executable size?

from this PR's modification, it added three strings
- `GNUmakefile'
- `makefile'
- `Makefile'

is this the reason?

1stDimension:
The issues with CI are not present in current master c697d066702ab81ce0684fedb4c638e0fc0473e8. I believe if you rebase your branch onto it the checks will pass. https://github.com/1stDimension/exa/pull/3 I tried to create similar environment in my fork.

ariasuni:

Issue #211, opened 2017-08-04
Should escape or quote file and directory names that can't be passed as is to the command line (like ls)


FSMaxB:
When you have a file or directory called `;` or `~` for example, you can't just delete them or do anything else with them without quoting. The same for some special characters. So what `ls` from coreutils does is to escape these characters or put the file names in quotes in the directory listing (so you could just copy paste the name and paste it into the command line if you want to do something with the file).

I think it would be useful to have that in exa as well.

jdanford:
GNU `ls` does a decent job of quoting such names, and it looks like the [go-shellquote](https://github.com/kballard/go-shellquote) package approximates its behavior.

Don42:
I couldn't find a way to get `ls` to quote special characters, but that might be because I'm using an old version.
I tried to build another escape function, although I'm not sure what the two styles are used for. So it might not work for all or even most usecases in exa.
[0] 510e2fde9aba40bc1c1b55d708de7c6b6e8d5764

HaleTom:
Here is an example of `ls` vs `exa`:

```
% /bin/ls -l --quoting-style=shell-escape
total 0
-rw-r--r-- 1 ravi ravi 0 Jul  5 16:56 ''$'\026'
drwxr-xr-x 1 ravi ravi 0 Jul  3 15:36  a
-rw-r--r-- 1 ravi ravi 0 Jul  3 11:10 'a b'
-rw-r--r-- 1 ravi ravi 0 Jul  3 11:12  bar
-rw-r--r-- 1 ravi ravi 0 Aug 11 18:14 'b c  '
-rw-r--r-- 1 ravi ravi 0 Jul  3 11:12  foo
-rw-r--r-- 1 ravi ravi 0 Jul  3 11:10 'x'$'\b''x'
% exa -l
.rw-r--r-- 0 ravi  5 Jul 16:56 \u{16}
drwxr-xr-x - ravi  3 Jul 15:36 a
.rw-r--r-- 0 ravi  3 Jul 11:10 a b
.rw-r--r-- 0 ravi  3 Jul 11:12 bar
.rw-r--r-- 0 ravi 11 Aug 18:14 b c
.rw-r--r-- 0 ravi  3 Jul 11:12 foo
.rw-r--r-- 0 ravi  3 Jul 11:10 x\u{8}x
```

flying-sheep:
I think modern shells also accept `$'x\bx'`, which would be a bit prettier than `'x'$'\b''x'`

[...]

Issue #885, opened 2021-05-29
Grid view doesn't work


Vistaus:
Grid view doesn't seem to work in 0.10.1 - it just displays a downward list. Even if I explicitly tell Exa to use e.g. 8 columns, it keeps displaying a downward list.

-Exa version: 0.10.1
-OS: Deepin 20.2 (based on Debian 10)
-Terminal Emulator: Deepin Terminal

Could you please look into this? If you need more information, please let me know.

Vistaus:
Actually, it does seem to work now, but only in a few select folders.

ariasuni:
Are you sure you’re not misusing [`COLUMNS`](https://github.com/ogham/exa/blob/master/man/exa.1.md#columns) (it doesn’t configure the number of colums in grid mode, but the terminal width used by exa)? If not, please give examples with the exact command and output, and the content of the relevant environment variables.

Vistaus:
I've never used the columns command before, so I don't think I'm misusing it.

I'm using exa -lGh

It seems to work in e.g. my home folder, but in other folders, it doesn't display any grid.

ariasuni:
OK but then what do you mean by: 

> Even if I explicitly tell Exa to use e.g. 8 columns

?

> It seems to work in e.g. my home folder, but in other folders, it doesn't display any grid.

Well I can’t reproduce, and my home folder is different from your home folder so I can’t attempt reproducing your issue…

Vistaus:
I meant export EXA_GRID_ROWS=xxx doesn't make it work.

ariasuni:
The purpose of `EXA_GRID_ROWS` is to disable the grid if the produced grid would have a number of rows inferior to the value of `EXA_GRID_ROWS`. Did you set it globally somewhere? That may the problem.

If not: it’s exhausting to ask you for details so please take the time to check and report the environment variables used to configure exa, and an example of incorrect output used by exa, this way I can attempt to reproduce your problem.

Vistaus:
I also tried without EXA_GRID_ROWS and the problem still exists. I installed Exa from the Debian repos, so the environment variables should be as expected. I did not alter them in any way.

I attached two screenshots, where you can see that it displays a grid in my home folder but not in my Downloads folder. (as an example)

![Schermfoto_20210529172357](https://user-images.githubusercontent.com/1716229/120075712-b2af4d00-c0a2-11eb-9637-928e0ae18a25.png)
[...]

Issue #318, opened 2017-11-27
Tests fail when not connected to a terminal


ignatenkobrain:
```
---- options::test::oneline_across stdout ----
	thread 'options::test::oneline_across' panicked at 'called `Result::unwrap_err()` on an `Ok` value: (Options { dir_action: List, filter: FileFilter { list_dirs_first: false, sort_field: Name(AaBbCc), reverse: false, dot_filter: JustFiles, ignore_patterns: IgnorePatterns { patterns: [] }, git_ignore: Off }, view: View { mode: Lines, colours: Colours { colourful: false, scale: false, filekinds: FileKinds { normal: Style {}, directory: Style {}, symlink: Style {}, pipe: Style {}, block_device: Style {}, char_device: Style {}, socket: Style {}, special: Style {}, executable: Style {} }, perms: Permissions { user_read: Style {}, user_write: Style {}, user_execute_file: Style {}, user_execute_other: Style {}, group_read: Style {}, group_write: Style {}, group_execute: Style {}, other_read: Style {}, other_write: Style {}, other_execute: Style {}, special_user_file: Style {}, special_other: Style {}, attribute: Style {} }, size: Size { numbers: Style {}, unit: Style {}, major: Style {}, minor: Style {}, scale_byte: Style {}, scale_kilo: Style {}, scale_mega: Style {}, scale_giga: Style {}, scale_huge: Style {} }, users: Users { user_you: Style {}, user_someone_else: Style {}, group_yours: Style {}, group_not_yours: Style {} }, links: Links { normal: Style {}, multi_link_file: Style {} }, git: Git { new: Style {}, modified: Style {}, deleted: Style {}, renamed: Style {}, typechange: Style {} }, punctuation: Style {}, date: Style {}, inode: Style {}, blocks: Style {}, header: Style {}, symlink_path: Style {}, broken_arrow: Style {}, broken_filename: Style {}, control_char: Style {} }, style: FileStyle { classify: JustFilenames, exts: NoFileColours } } }, [])', src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
---- options::view::test::views::across stdout ----
	Testing Ok(Lines)
thread 'options::view::test::views::across' panicked at 'assertion failed: false', src/options/view.rs:526:8
---- options::view::test::views::empty stdout ----
	Testing Ok(Lines)
thread 'options::view::test::views::empty' panicked at 'assertion failed: false', src/options/view.rs:521:8
[...]

PR #1165, opened 2023-02-24
Quote file names containing spaces and/or single quotes

### LINKED ISSUE #216: no indication that filename ends in space
When a filename ends in a space, there is no indication. Compare this with GNU ls, which encloses filenames with spaces in quotes.

```sh
$ touch hello
$ ls
hello
$ exa
hello
$ touch 'hello '
$ ls
 hello	'hello '
$ exa
hello   hello
```

As an aside on the example above, GNU ls sorts `'hello '` after `'hello'` while exa sorts them the other way round.

DenebTM:
If a filename contains spaces, it will be wrapped in single quotes ('). If a filename contains literal single quotes, it will instead be wrapped in double quotes ("). This somewhat mimicks the default behaviour of modern-day GNU `ls`.

This resolves #216 and contributes to #211. It also cleans up the code in the non-printable characters loop a bit. I personally believe quotes should be the default, but a command-line switch (`--no-quotes`) exists for those who prefer not having them.

My intent with this PR is to resolve something that was bothering me personally, with minimal code changes, in a way that is satisfactory to myself. I believe it presents an improvement over the status quo, and further improvements may be the subject of future PRs.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,7 +35,7 @@ use crate::fs::{Dir, File};
 use crate::fs::feature::git::GitCache;
 use crate::fs::filter::GitIgnore;
 use crate::options::{Options, Vars, vars, OptionsResult};
-use crate::output::{escape, lines, grid, grid_details, details, View, Mode};
+use crate::output::{escape, lines, grid, grid_details, details, View, Mode, file_name};
 use crate::theme::Theme;
 
 mod fs;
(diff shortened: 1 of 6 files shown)


DenebTM:
Changed to draft because I've realised that this messes with multi-column display in its current state.

DenebTM:
After some more testing, I've realised that the grid view is currently broken on `main` - see recent comments on #885.

Reopening as the problem lies somewhere else, not with this PR.

DenebTM:
closing this as the changes have been merged into the maintained fork, [eza](https://github.com/eza-community/eza) with [#318](https://github.com/eza-community/eza/pull/318)

Issue #396, opened 2018-06-13
Add support for rendering filenames as hyperlinks


liff:
Some terminal emulators [support escape sequences that can be used to mark text as a hyperlink](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).

Could a similar option perhaps be added to `exa`? Coreutils `ls` [supports this](http://git.savannah.gnu.org/cgit/coreutils.git/commit/?id=799bac0d06cfabe9491498727308df8d1aca6d98) since 8.28.

Don42:
Sounds interesting, I expect this depends on #211 being implemented.

ogham:
I had no idea about this feature, but it looks really useful!

jcklpe:
+1 to this feature.

ahsanp:
Hi! Given that ansi_term doesn't yet support hyperlink painting, would it be okay to make a pull request that supports this painting without using ansi_term? 
More specifically, one would add in the hyperlink escape codes in the render functions. That way we can avoid messing up text cell width in the output.

ariasuni:
Well, I guess you could just make a PR for [`ansi_term`](https://github.com/ogham/rust-ansi-term) (which is also maintained by ogham) directly, though it sounds a little bit more complicated.

ahsanp:
That sounds great! Thank you!

mpldr:
> No longer sure if there really is a dependency. I forgot how complex character
> escapes can be.

Pretty easy actually… since it's just a URI, you can use normal URI-encoding (space -> %20 and so on). Maybe this helps (it's in Go though) https://git.sr.ht/~poldi1405/go-ansi/tree/master/item/hyperlinks.go#L20

XhstormR:
This feature is really helpful.

unrelentingtech:
Looks like there is an ansi_term PR: https://github.com/ogham/rust-ansi-term/pull/61

phorcys420:
le 2018

ariasuni:
Well, I didn’t look at termwiz, but it could be interesting to try to use it instead of rust-ansi-term and see if it’s still as performant/doesn’t grow the binary too much/etc.

aaccioly:
Kitty has a great implementation of Hyperlink actions. To the point that I now naturally expect to be able to click on files, URLs, etc and expect my terminal emulator to launch something useful.
It would be lovely if exa would implement this feature.

mpldr:
[...]

PR #1177, opened 2023-04-12
feature: hyperlink flag

dankeyy:
adding a `--hyperlink` option like in `ls`
feel free to throw any comments at the code

Edit: it seems like your inline output is broken on master (is fine on latest release), 
meaning even if output can be displayed on one-two lines, it will spread as one file per line (try running exa from master on the exa root dir, then try it the same with v0.10.1/ ls)

As far as I've tested, this is because of the upgrade to term_grid 0.2.0
since then, `grid.fit_into_width(self.console_width)` returns None for what would've been valid `console_width` in past versions.

I will at least match the behavior of hyperlink to regular exa, so that if/ when you do fix it, it'll work fine with the hyperlink flag aswell

also read my comments, especially on the new escape function

thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ exa’s options are almost, but not quite, entirely unlike `ls`’s.
 - **--colo[u]r-scale**: highlight levels of file sizes distinctly
 - **--icons**: display icons
 - **--no-icons**: don't display icons (always overrides --icons)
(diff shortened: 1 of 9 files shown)


nonZero:
Refs #396

ariasuni:

PR #1178, opened 2023-04-12
Replace obsolete icons

mtoohey31:
This merge request replaces icons that have been relocated whose old location has been marked as obsolete in nerd-fonts.

The tldr, as I understand it, is that the codepoints used by mdi characters were relocated around the start of 2022 because the original positions conflicted with certain Chinese characters, so they were moved in [this nerd-fonts merge request](https://github.com/ryanoasis/nerd-fonts/pull/773). This merge request updates exa's hard-coded icon values to use the new codepoints. This change should be transparent to users as long as they've updated their font sometime in the last year.

The only difference I'm noticed wasa slight difference between the new and old versions of the xml icon. On kitty, the old icon takes up two cells, messing up the filename's alignment:

2023-04-12T19:14:29,591563750-04:00

The new icon doesn't doesn't seem to do this:

2023-04-12T19:14:20,143546614-04:00

So I think this is an improvement, even though I didn't expect to see any change?

### DIFF
--- a/src/output/icons.rs
+++ b/src/output/icons.rs
@@ -140,13 +140,13 @@ pub fn icon_for_file(file: &File<'_>) -> char {
             "cp"            => '\u{e61d}', // 
             "cpio"          => '\u{f410}', // 
             "cpp"           => '\u{e61d}', // 
-            "cs"            => '\u{f81a}', // 
+            "cs"            => '\u{f031b}', // 󰌛
             "csh"           => '\u{f489}', // 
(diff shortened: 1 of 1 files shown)


mtoohey31:
Looks like the difference I was talking about at the end there is caused by this: https://github.com/ryanoasis/nerd-fonts/issues/1061. Like I mentioned above though, I think this change is in our favour since the smaller icon prevents the columns from becoming misaligned.

PR #1202, opened 2023-06-11
feat: Add a time style with better human readability

niT-Tin:
Because when using the built-in command ls of nushell, I found that it has a more readable date display, so I hope that Exa can also support it
The following is the result of running ls in a nushell
nushell
Here are the results of running exa
exa

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,8 @@ term_grid = "0.2.0"
 terminal_size = "0.1.16"
 unicode-width = "0.1"
 zoneinfo_compiled = "0.5.1"
+chrono = "0.4.26"
+chrono-humanize = "0.2.2"
 
 [target.'cfg(unix)'.dependencies]
 users = "0.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ name = "exa"
 
 [dependencies]
 ansi_term = "0.12"
+chrono = "0.4.26"
+chrono-humanize = "0.2.2"
 glob = "0.3"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -32,6 +32,8 @@ term_grid = "0.2.0"
 terminal_size = "0.1.16"
 unicode-width = "0.1"
 zoneinfo_compiled = "0.5.1"
+chrono = "0.4.26"
```
planet36, line 35: It looks like the dependencies list should be sorted alphabetically.

src/options/view.rs
```diff
@@ -266,6 +266,9 @@ impl TimeFormat {
         else if &word == "full-iso" {
             Ok(Self::FullISO)
         }
+        else if &word == "hu" {
```
planet36, line 269: Why `hu` instead of `human` or `relative`?

planet36:
This would be a good feature.  The `--date=relative` option for lsd is the only reason I use lsd instead of exa.

src/options/view.rs
niT-Tin: Yes! I also thought that after seeing nushell, exa should have come up with the idea of this PR, which was a temporary idea, so I didn't think much about it. However, now It's better to use "human" or "relative" instead.

planet36 (approved):
Nice work!

ariasuni:
Closing this since exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and this has been merged in [the active fork eza](https://github.com/eza-community/eza). Thanks!

PR #1204, opened 2023-06-18
Support time zone settings for Windows

CID8705:
The time zone setting for Windows was unsupported, so it is now supported.
In making this correction, a new crate chrono was added.
Also, the difference appears in Cargo.lock, but since Cargo.lock is generated automatically, so I added it to .gitignore.
Reference: https://github.com/github/gitignore/blob/main/Rust.gitignore


ariasuni:
You shouldn’t remove `Cargo.lock` from the repository, [see here for a rationale](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

The code seems quite complicated and I’ve honestly no idea what it does, do you think you could simplify and explain it? I don’t think we need to use `Arc` for timezones in exa for example.

CID8705:
I see why Cargo.lock should not be removed.
I would like to correct this.

We are just modifying from TimeZoneSource::Static() to TimeZoneSource::Runtime() and setting the appropriate time zone.
The basic structure of the source code has not changed much from the original one.
the reason for using Arc is to match the type of the original source code.
The source code of zoneinfo_compiled was used as a reference.

Reference: https://github.com/rust-datetime/zoneinfo-compiled/blob/master/src/lib.rs#L44

PR #1228, opened 2023-08-04
Respect GIT_CEILING_DIRECTORIES

### LINKED ISSUE #1226: `GIT_CEILING_DIRECTORIES` is not respected
# Problem

When running a command with the `--git` option, it seems that `GIT_CEILING_DIRECTORIES` is not respected.

This can lead to huge delays when `exa` looks past a ceiling directory, finds a `.git` in some larger parent, and checks that entire larger tree for Git changes.

# Reproducible Example

The easiest repro is probably to temporarily make your home directory into a git repo, and see what that does to your execution time, with or without `GIT_CEILING_DIRECTORIES=~` in the environment:

```sh
mkdir ~/test &&
cd ~/test &&
time exa --all --long --git  # should be fast

cd ~ &&
git init &&
git commit --allow-empty -m 'test' &&
cd ~/test &&
time exa --all --long --git  # should be slow

export GIT_CEILING_DIRECTORIES=~ &&
time exa --all --long --git  # should be fast again
```

This is great for a test because your home directory is probably big enough to see the slowdown. (Just don't forget to `rm -rf ~/.git` when done, unless you're into that sort of thing... if you've just discovered you might be, I recommend `git config status.showUntrackedFiles no` and `export GIT_CEILING_DIRECTORIES=~`.)

# Misc.

Output of `exa --version` (but I would guess this happens with any version):

```
v0.10.1 [+git]
https://the.exa.website/
```

Observed on the following OS+hardware combinations (but I would guess it this happens on any OS and any hardware):

1. Debian Linux on x86-64 CPU (Librem 14).
2. Debian in WSL2 in Windows 11 on x86-64 CPU (different laptop).
3. Termux in Android on Aarch64 CPU (two different phone models).

mentalisttraceur:
Fixes #1226.

### DIFF
--- a/src/fs/feature/git.rs
+++ b/src/fs/feature/git.rs
@@ -162,7 +162,9 @@ impl GitRepo {
     /// Returns the original buffer if none is found.
     fn discover(path: PathBuf) -> Result<Self, PathBuf> {
         info!("Searching for Git repository above {:?}", path);
-        let repo = match git2::Repository::discover(&path) {
+        let flags = git2::RepositoryOpenFlags::FROM_ENV;
+        let unused: [&OsStr; 0] = [];
+        let repo = match git2::Repository::open_ext(&path, flags, &unused) {
             Ok(r) => r,
             Err(e) => {
                 error!("Error discovering Git repositories: {:?}", e);


mentalisttraceur:
What? Deleting my fork closes the PR now? And there's no way to reopen it without recreating! Very annoying. I don't like keeping forks sitting around on my account indefinitely.

Anyway, when someone's ready to do anything with this PR, let me know and I'm happy to re-create it.

PR #1221, opened 2023-07-24
add icon support for .mjs, .cjs, .mts, .cts files 

aarondill:
These are javascript/typescript files that specify the module type (in NodeJS). Thusly, they should have icons matching their language (JS/TS).

PR #1194, opened 2023-05-16
add Emacs icon for .el and org-mode for .org

lb-wilson:
Hello! Hoping to add icons for Emacs org-mode files (.org) and emacs-lisp source files (.el). These are common extensions, and the icons are included as part of Nerd Fonts icons and should not conflict with any other icons. Thank you!

PR #1150, opened 2022-12-24
Add MP2 audio format icon

rilysh:
This PR adds MP2 audio file extension and the icon. For more information about MP2 (MPEG-1), see [here](https://en.wikipedia.org/wiki/MPEG-1_Audio_Layer_II)
Official standard webpage: [here](https://mpeg.chiariglione.org/standards/mpeg-1/audio)

PR #1148, opened 2022-12-19
Fixed issue #1106: added icon for '.gpg' files

marcoradocchia:
Added lock icon ('ea75') for files with '.gpg' extension, fixing issue #1106.
Below a test output:
```sh
~/gpg_icon
λ exa --icons
 test.gpg
```

PR #1229, opened 2023-08-04
feat: Use perfect hash tables for file types and icons

cfxegbert:
Change filetype and icon lookup to use perfect hash tables to simplify lookups.

- Perfect hash maps are generated at compile instead of runtime avoiding the lazy_static initialization of MAP_BY_NAME.  Right now the tables are directly defined in build.rs but could be easily moved to an external datafile that is read by build.rs
- Perfect hash maps are sized to the data and do not need to allocate extra buckets that a regular hash map creates.
- Avoid the linear searches for the file extensions in both the filetype and icon.  Match statements of str does a linear walk of each pattern.
- Remove the coupling between icon and filetype.  This will allow a more specific icon to be used instead of the generic icon for filetype.  For example, a file with an svg extension can use the filetype color for an image but use an icon to shows it is vector art instead of a bitmap image.  As a result of removing the coupling all the filetypes for image, video, music, and lossless had to be added to the icon map.  Most of them were already in the icon map.
- Create constants for a majority of unicode icons. 
- Fix icons removed in Nerd Fonts 3.x


cfxegbert:
Added pull request #1221

cfxegbert:
Added pull request #1194 #1191 #1162 #1150 #1148 #1144 #1140 #1120 #1100 #1238 #1234

cfxegbert:
Matched icons with https://github.com/eza-community/eza

src/info/filetype.rs
```diff
@@ -5,140 +5,74 @@
...
-    fn is_temp(&self, file: &File<'_>) -> bool {
-        file.name.ends_with('~')
-            || (file.name.starts_with('#') && file.name.ends_with('#'))
-            || file.extension_is_one_of( &[ "tmp", "swp", "swo", "swn", "bak", "bkp", "bk" ])
-    }
-
-    fn is_compiled(&self, file: &File<'_>) -> bool {
-        if file.extension_is_one_of( &[ "class", "elc", "hi", "o", "pyc", "zwc", "ko" ]) {
-            true
+    fn get_file_type(file: &File<'_>) -> Option<FileType> {
```
cafkafk, line 37: This should have some doc comments

build.rs
```diff
@@ -42,9 +50,26 @@ fn main() -> io::Result<()> {
...
+fn create_file_typing_hash_file() -> io::Result<()> {
+    let path = &PathBuf::from(env::var("OUT_DIR").unwrap()).join("filetype_maps.rs");
+    let mut file = io::BufWriter::new(File::create(path).unwrap_or_else(|_| { panic!("{}", path.to_string_lossy().to_string()) }));
+    generate_extension_type_map(file.get_mut())?;
+    generate_filename_type_map(file.get_mut())?;
+    file.flush()
+}
 
-    Ok(())
+/// Create the perfect hashing for file icons
```
cafkafk, line 65: I think you should expand on this, how it's done and the motivation for doing it this way.

build.rs
```diff
@@ -121,3 +147,791 @@ fn build_date() -> String {
...
+                 .entry("localtime",          "'\u{f43a}'")           // 
+                 .entry("robots.txt",         "'\u{f06a9}'")          // 󰚩
+                 .entry("timezone",           "'\u{f43a}'")           // 
+                 .entry("webpack.config.js",  "'\u{f072b}'")          // 󰜫
+                 .entry("yarn.lock",          "'\u{e6a7}'")           // 
+                 .build()
+    )
+}
+
+fn generate_directory_icon_map(file: &mut File) -> io::Result<()> {
```
cafkafk, line 892: This should have some doc comments

cafkafk (changes requested):

cafkafk:
ohh wrong pr

ariasuni:
exa is unmaintained, and this has been merged in [the active fork eza](https://github.com/eza-community/eza), so I see no reason to keep this PR around (thanks a lot for the effort though). Also see https://github.com/ogham/exa/issues/1243

Issue #1225, opened 2023-08-01
Would you be interested in having a Dockerfile for development?


AgustinRamiroDiaz:
I could work on a dockerfile/docker-compose to isolate development if it's something you'd like in the repo :)

Maybe it's also useful to create releases to different targets :thinking: 

We can even do a dev-container for vscode (https://code.visualstudio.com/docs/devcontainers/containers)

ariasuni:
Well in general I don’t see the point, but in exa we use Vagrant for tests involving specific directory structures, locales, etc. I wanted to transition the Vagrantfile to a Dockerfile to be able to run these «xtests» that already exist, but I didn’t have the time. That would be great to have that!

AgustinRamiroDiaz:
Awesome! I'll work on it :D

AgustinRamiroDiaz:
Hey @ariasuni!

quick question: are Vagrant tests passing with the latest master commit? 

I don't see any CI with Vagrant to check this, and couldn't make Vagrant work in my PC :(

I have been able to do a PoC of the Dockerfile and I could execute the tests, but there's a lot of formatting errors. I want to know if this errors are a misconfiguration of mine or if they are actually failing in master

ariasuni:
Closing this since exa is unmaintained (see https://github.com/ogham/exa/issues/1243), you should take a look at [the active fork eza](https://github.com/eza-community/eza) (an alternative strategy has been used for tests).

PR #1230, opened 2023-08-07
Add Dockerfile to enable running end-to-end tests with docker

AgustinRamiroDiaz:
# Context
Fixes https://github.com/ogham/exa/issues/1225

# Reviewer tips
The commit history is a bit dirty (if we go forward we should squash it), so look at the final picture
- Added `Dockerfile` and `.dockerignore`
- Documented a bit in the `README.md`

I also encourage you to run the tests in your local machine to see how it works. Documentation on how to do this in the `README.md`

# Notes
- This is my first time contributing here, I'm open to receive all your feedback
- The tests are failing at the moment in the tip of `master`. If someone is able to run vagrant, please notify me if this is also the case so I know if it's an issue with the `Dockerfile` configuration

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,12 +18,10 @@ RUN cargo install -q cargo-hack
 RUN cargo install -q --git https://github.com/ogham/specsheet
 
 # Install Just, the command runner.
-RUN wget -q "https://github.com/casey/just/releases/download/v0.8.3/just-v0.8.3-x86_64-unknown-linux-musl.tar.gz"
-RUN tar -xf "just-v0.8.3-x86_64-unknown-linux-musl.tar.gz"
-RUN cp just /usr/local/bin
+RUN curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin
 
-# Guarantee that the timezone is UTC — some of the tests depend on this (for now).
-RUN timedatectl set-timezone UTC
+# TODO: Guarantee that the timezone is UTC — some of the tests depend on this (for now).
+# RUN timedatectl set-timezone UTC
 
 ARG DEVELOPER=root
 # Use a different ‘target’ directory on the VM than on the host.
@@ -50,16 +48,18 @@ RUN ln -sf /usr/bin/run-xtests /usr/bin/x
 RUN echo -e "#!/bin/sh\nbuild-exa && test-exa && run-xtests" > /usr/bin/compile-exa
 RUN ln -sf /usr/bin/compile-exa /usr/bin/c
 
+ADD --chmod=+x devtools/dev-package-for-linux.sh /vagrant/devtools/dev-package-for-linux.sh
 RUN echo -e "#!/bin/sh\nbash /vagrant/devtools/dev-package-for-linux.sh \\$@" > /usr/bin/package-exa
 RUN echo -e "#!/bin/sh\ncat /etc/motd" > /usr/bin/halp
 
-RUN chmod +x /usr/bin/{exa,rexa,b,t,x,c,build-exa,test-exa,run-xtests,compile-exa,package-exa,halp}
+# RUN chmod +x /usr/bin/{exa,rexa,b,t,x,c,build-exa,test-exa,run-xtests,compile-exa,package-exa,halp}
 
 
 # Configure the welcoming text that gets shown:
 
 # Capture the help text so it gets displayed first
 RUN rm -f /etc/update-motd.d/*
+ADD --chmod=+x devtools/dev-help.sh /vagrant/devtools/dev-help.sh
 RUN bash /vagrant/devtools/dev-help.sh > /etc/motd
 
 # Tell bash to execute a bunch of stuff when a session starts
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,14 +106,14 @@ RUN bash /vagrant/devtools/dev-set-up-environment.sh
 RUN bash /vagrant/devtools/dev-create-test-filesystem.sh
 
 WORKDIR /vagrant
-COPY --from=exa /app/target /vagrant/target
+COPY --link --from=exa /app/target /vagrant/target
 
 # TODO: remove this once tests don't depend on it
 RUN ln -s /vagrant/* ${HOME}
 
-COPY --from=specsheet /usr/local/cargo/bin/specsheet /usr/bin/specsheet
-COPY --from=cargo-hack /usr/local/cargo/bin/cargo-hack /usr/bin/cargo-hack
-COPY --from=just /usr/bin/just /usr/bin/just
+COPY --link --from=specsheet /usr/local/cargo/bin/specsheet /usr/bin/specsheet
+COPY --link --from=cargo-hack /usr/local/cargo/bin/cargo-hack /usr/bin/cargo-hack
+COPY --link --from=just /usr/bin/just /usr/bin/just
 
 FROM base AS test
 CMD ["/vagrant/xtests/run.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,19 @@
-FROM rust:1.71.1 AS just
+ARG base-rust=rust:1.71.1
+
+FROM base-rust AS just
 # Install Just, the command runner.
 RUN curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/bin
 
-FROM rust:1.71.1 AS specsheet
+FROM base-rust AS specsheet
 RUN cargo install -q --git https://github.com/ogham/specsheet
 
-FROM rust:1.71.1 AS cargo-hack
+FROM base-rust AS cargo-hack
 RUN cargo install -q cargo-hack
 
-FROM rust:1.71.1 AS cargo-kcov
+FROM base-rust AS cargo-kcov
 RUN cargo install -q cargo-kcov
 
-FROM rust:1.71.1 AS exa
+FROM base-rust AS exa
 WORKDIR /app
 # Copy the source code into the image
 COPY . .
@@ -21,7 +23,7 @@ RUN cargo build
 # We use Ubuntu instead of Debian because the image comes with two-way
 # shared folder support by default.
 # This image is based from Ubuntu
-FROM rust:1.71.1 AS base
+FROM base-rust AS base
 
 # Install the dependencies needed for exa to build
 RUN apt-get update -qq
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,11 @@ RUN apt-get install -y \
 RUN ln -s $(which python3) /usr/bin/python
 RUN cargo kcov --print-install-kcov-sh | sh
 
-ADD --chmod=+x devtools/* /vagrant/devtools/
+COPY --chmod=+x . /vagrant/
+
+# Make sudo dummy replacement, so we don't weaken docker security
+RUN echo -e "#!/bin/bash\n\$@" > /usr/bin/sudo
+RUN chmod +x /usr/bin/sudo
 
 RUN bash /vagrant/devtools/dev-set-up-environment.sh
 RUN bash /vagrant/devtools/dev-create-test-filesystem.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,25 +51,27 @@ RUN ln -s $(which python3) /usr/bin/python
 RUN cargo kcov --print-install-kcov-sh | sh
 
 # Create a variety of misc scripts.
+RUN <<EOF
+  ln -sf /vagrant/devtools/dev-run-debug.sh /usr/bin/exa
+  ln -sf /vagrant/devtools/dev-run-release.sh /usr/bin/rexa
 
-RUN ln -sf /vagrant/devtools/dev-run-debug.sh /usr/bin/exa
-RUN ln -sf /vagrant/devtools/dev-run-release.sh /usr/bin/rexa
+  echo -e "#!/bin/sh\ncargo build --manifest-path /vagrant/Cargo.toml \\$@" > /usr/bin/build-exa
+  ln -sf /usr/bin/build-exa /usr/bin/b
 
-RUN echo -e "#!/bin/sh\ncargo build --manifest-path /vagrant/Cargo.toml \\$@" > /usr/bin/build-exa
-RUN ln -sf /usr/bin/build-exa /usr/bin/b
+  echo -e "#!/bin/sh\ncargo test --manifest-path /vagrant/Cargo.toml \\$@ -- --quiet" > /usr/bin/test-exa
+  ln -sf /usr/bin/test-exa /usr/bin/t
 
-RUN echo -e "#!/bin/sh\ncargo test --manifest-path /vagrant/Cargo.toml \\$@ -- --quiet" > /usr/bin/test-exa
-RUN ln -sf /usr/bin/test-exa /usr/bin/t
+  echo -e "#!/bin/sh\n/vagrant/xtests/run.sh" > /usr/bin/run-xtests
+  ln -sf /usr/bin/run-xtests /usr/bin/x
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+RUN ln -s /vagrant/completions/zsh/_exa /usr/share/zsh/vendor-completions/_exa
+
+# fish
+RUN ln -s /vagrant/completions/fish/exa.fish /usr/share/fish/completions/exa.fish
+
+
+# Copy the source code into the image
+COPY --chmod=+x . /vagrant/
+
+# Make sudo dummy replacement, so we don't weaken docker security
```
mentalisttraceur, line 101: So... I think there's some misunderstanding here, because Docker containers are implemented with Linux namespaces, which nowadays includes even a separate user namespace, so containers have their own root which is separate from (and has no privileges in) the main system.

So `sudo` in a container *can't* weaken the security of anything outside that container. (And inside the container, the only thing that coud be weakened is if we were running an app whose security depended on things running with privileges dropped even within the container, but that doesn't seem to be the case for these tests.)

mentalisttraceur (commented):
Nice! Thanks for doing this!

I hope it gets merged (once everything is verified working as-it's-supposed-to) because I also couldn't get the in-Vagrant tests running, and it might be nicer to have both as options (or whichever single option is most widely available if Exa doesn't want to maintain both).

In the big picture, the only concern I have is that if the VM tests currently exercise listing things belonging to different users/groups, we should probably make sure the Docker container is able to do the same.

A few detailed/small/nitpicky comments inline. Overall great contribution though, in my opinion!

README.md
```diff
@@ -215,27 +213,45 @@ Once Rust is installed, you can compile exa with Cargo:
...
 
 - If you intend to compile for musl, you will need to use the flag `vendored-openssl` if you want to get the Git feature working.
-The full command is `cargo build --release --target=x86_64-unknown-linux-musl --features vendored-openssl,git`.
+  The full command is `cargo build --release --target=x86_64-unknown-linux-musl --features vendored-openssl,git`.
 
 For more information, see the [Building from Source page](https://the.exa.website/install/source).
 
+### End to end testing
+
+## Testing with containers
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: Nice catch!

README.md
```diff
@@ -247,7 +263,7 @@ The initial attempt to solve the problem was just to create a directory of “aw
 But even this output would change if, say, the user’s locale formats dates in a different way.
 These can be mocked inside the code, but at the cost of making that code more complicated to read and understand.
 
-An alternative solution is to fake *everything*: create a virtual machine with a known state and run the tests on *that*.
+An alternative solution is to fake _everything_: create a virtual machine with a known state and run the tests on _that_.
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: Yeah that's true. It was my IDE that automatically formatted the code. Rolled backed!

Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+ADD --chmod=+x devtools/dev-help.sh /vagrant/devtools/dev-help.sh
+RUN bash /vagrant/devtools/dev-help.sh > /etc/motd
+
+# Tell bash to execute a bunch of stuff when a session starts
+RUN echo "source /vagrant/devtools/dev-bash.sh" > ${HOME}/.bash_profile
+
+# Disable last login date in sshd
+# RUN sed -i '/PrintLastLog yes/c\PrintLastLog no' /etc/ssh/sshd_config
+# RUN systemctl restart sshd
+
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: I was saving it just in case it had some relationship with why the tests didn't pass, but yeah I'll remove them until we figure that out

Dockerfile
AgustinRamiroDiaz: Yeah that's true. I'm not gonna lie, I copy pasted it from stack overflow :sweat_smile: 

Great explanation and thanks for the feedback!

I'm fixing the comment :D

polarathene (commented):
I'm a bit busy this week, but I could contribute a review with feedback for changes if you ping me next week if you'd like.

---

Your `.dockerignore` is fairly extensive, if you don't need as much in the image, you can ignore everything and then add some exceptions of folders / files to include in the docker context. See [this example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/.dockerignore) for reference.

I help maintain a Docker focused project so `Dockerfile` is something I know quite well. Assuming anyone that is building the image is ok with doing so with a fairly recent version of Docker (v23+, and depending on features versions released in 2022), there is a few features you could take advantage of.

- HereDoc for many batching commands into a single layer without the prior drawback of chaining `&& \` at the end of each line. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/Dockerfile#L23-L26).
- It's ok to extract some scripts out to their own shell script file if appropriate, sometimes it's just noise to the Dockerfile and doesn't belong there, better linting too. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/Dockerfile#L28-L31).
- You may want to use `--link` with your `COPY` instructions, it can provide an optimization that minimizes invalidating layers. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/Dockerfile#L67). Might be good for the series of `--from` stage copies you have at the end.
- You create a few shell files (`t`, `rexa`, etc) to run different commands, but could use a single shell script with a switch to support that. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/bin/fail2ban#L6-L10) (not the best).
- I think all the `rust:1.71.1` could reference a single `ARG` with default value instead.
- `apt-get` could use a cache mount if the image was built often enough and you wanted to minimize this step. From the looks of it the cache mount may not be too useful here, but it is for anything with longer build times that can leverage cache from prior builds. This is different from `docker build` leveraging cached layers and is mostly beneficial for parts of a build that are not needed in the image like package cache, build cache etc.

AgustinRamiroDiaz:
@polarathene thanks a lot for your feedback! I'll be addressing it in my next commits

> chaining && \

Where do you think we could use this feature? We don't have the command chaining, we only have some multi-line commands only for readability (install a lot of apt packages)

> It's ok to extract some scripts out to their own shell script file if appropriate, sometimes it's just noise to the Dockerfile and doesn't belong there, better linting too. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/Dockerfile#L28-L31).

Yeah I agree, but for this PR I don't want to be too disruptive, I think this is great as a post iterative improvement

> You create a few shell files (t, rexa, etc) to run different commands, but could use a single shell script with a switch to support that. [Example](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/bin/fail2ban#L6-L10) (not the best).

Sorry, I didn't understand the example. If you mean to modify the current scripts, I do not want to modify the internal behavior because I cannot guarantee that they'll behave correctly in Vagrant (couldn't make it work)

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+# zsh
+RUN ln -s /vagrant/completions/zsh/_exa /usr/share/zsh/vendor-completions/_exa
+
+# fish
+RUN ln -s /vagrant/completions/fish/exa.fish /usr/share/fish/completions/exa.fish
+
+
+# Copy the source code into the image
+COPY --chmod=+x . /vagrant/
```
polarathene, line 98: This seems like it should have been done prior to the scripts above relying on the contents? Probably why you also were using `ADD` for some scripts earlier.

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+
+# Make sudo dummy replacement
+# This is needed for some tests that use sudo
+RUN echo -e '#!/bin/sh\n"$@"' > /usr/bin/sudo
+RUN chmod +x /usr/bin/sudo
```
polarathene, line 103: Really shouldn't be necessary to rely on sudo if the intent is to replace Vagrantfile with Dockerfile.

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+
+WORKDIR /vagrant
+COPY --from=exa /app/target /vagrant/target
+
+# TODO: remove this once tests don't depend on it
+RUN ln -s /vagrant/* ${HOME}
+
+COPY --from=specsheet /usr/local/cargo/bin/specsheet /usr/bin/specsheet
+COPY --from=cargo-hack /usr/local/cargo/bin/cargo-hack /usr/bin/cargo-hack
+COPY --from=just /usr/bin/just /usr/bin/just
```
polarathene: kconv should be located down here with the rest? I assume the concern was due to the package install portion for kconv and moving the layer(s) down to the end here being problematic when an earlier instruction invalidates the `Dockerfile` cache?

polarathene (changes requested):

polarathene:
> Where do you think we could use this feature? We don't have the command chaining, we only have some multi-line commands only for readability (install a lot of apt packages)

I see now that you've mostly taken the scripts from the `Vagrantfile` as-is. I provided some suggestions of where you could use it.

If both files are to be kept, it probably should be it's own script file that is shared between the two. Personally I'd be in favor of removing the `Vagrantfile` and simplify the `Dockerfile.

> Yeah I agree, but for this PR I don't want to be too disruptive, I think this is great as a post iterative improvement
> I do not want to modify the internal behavior because I cannot guarantee that they'll behave correctly in Vagrant (couldn't make it work)

I haven't looked too much into how those were being used, but since this is just taking lines from the `Vagrantfile` I can understand why you want to take that approach.

Dockerfile
polarathene: You could actually shift these up to after kconv, shouldn't be an issue given what the rest of the stage is doing?

Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+# Copy the source code into the image
+COPY --chmod=+x . /vagrant/
+
+# Make sudo dummy replacement
+# This is needed for some tests that use sudo
+RUN echo -e '#!/bin/sh\n"$@"' > /usr/bin/sudo
+RUN chmod +x /usr/bin/sudo
+
+RUN bash /vagrant/devtools/dev-set-up-environment.sh
+RUN bash /vagrant/devtools/dev-create-test-filesystem.sh
```
polarathene, line 106: I'm not sure if you can use `mknod` with the image build to get the `/dev` special files created:

https://github.com/ogham/exa/blob/f039202b4f508200739e2114f74a539344d5bd1c/devtools/dev-create-test-filesystem.sh#L126-L128

Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+
+# Copy the source code into the image
+COPY --chmod=+x . /vagrant/
+
+# Make sudo dummy replacement
+# This is needed for some tests that use sudo
+RUN echo -e '#!/bin/sh\n"$@"' > /usr/bin/sudo
+RUN chmod +x /usr/bin/sudo
+
+RUN bash /vagrant/devtools/dev-set-up-environment.sh
```
polarathene, line 105: <details>
<summary>Ignore this</summary>

This also has some odd locale logic at the end. I'm wondering if that's necessary?

In [`docker-mailserver` image](https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/Dockerfile#L275-L280) we have a similar locale section:

```Dockerfile
RUN <<EOF
  rm -rf /usr/share/locale/*
  rm -rf /usr/share/man/*
  rm -rf /usr/share/doc/*
  update-locale
EOF
```

For exa, I tracked it down to a specific test for different locales, so ignore this feedback 😅 

https://github.com/ogham/exa/commit/115315a03cefbdc6ec0da7ab174113a5950434f3

</details>

Dockerfile
AgustinRamiroDiaz: I agree, I just don't want to touch all the internal files because a) I'm not too familiar with them and b) I'm not sure it'll behave the same as I couldn't run the Vagrant tests. I think we should go iteratively on this, by first making it compatible and afterwards refactoring the `sudo` stuff

Dockerfile
AgustinRamiroDiaz: Yeah makes more sense. The symlinks should still work though, but I think it's better to do it as you mention. Changed in https://github.com/ogham/exa/pull/1230/commits/59e9c47674730cd6145146aedd64931fb95b2103

Dockerfile
AgustinRamiroDiaz: > kconv should be located down here with the rest?

Actually it's needed for `RUN cargo kcov --print-install-kcov-sh | sh` [here](https://github.com/AgustinRamiroDiaz/exa/blob/bb31f84f9f24dcacebac8edc4891fc3ac6175a5b/Dockerfile#L51)

> You could actually shift these up to after kconv, shouldn't be an issue given what the rest of the stage is doing?

That's true, do you think it's useful? I don't understand the benefit

Dockerfile
AgustinRamiroDiaz: Sorry, I don't understand. Do you mean that `mknod` won't work in docker?

AgustinRamiroDiaz:
@polarathene @mentalisttraceur thanks a lot for your time and feedback. I think I've addressed it all, please ping me if not

Dockerfile
polarathene: Yeah, I don't think it works, but I could be mistaken, I haven't checked 🤷‍♂️

Dockerfile
polarathene: >I don't understand the benefit

The layers that are more frequently to change should come later. The bulk of this stage is something that could run in a separate script via `RUN`, but for now you've stated you want to keep it in the `Dockerfile`.

You could move these `COPY` lines up before `kconv` for better co-location, but right now no major concern either way.

README.md
```diff
@@ -234,9 +234,26 @@ The full command is `cargo build --release --target=x86_64-unknown-linux-musl --
...
 For more information, see the [Building from Source page](https://the.exa.website/install/source).
 
+### End to end testing
 
-### Testing with Vagrant
+#### Testing with containers
 
+You can use [Docker](https://www.docker.com/) to run exa in a container, and test it against a known state.
+
+```fish
```
polarathene, line 243:
````suggestion
```sh
````

polarathene:
Thanks for addressing the feedback :)

I'm quite busy so might not have much more feedback for now, especially since further work is being deferred to a future PR.

I haven't tried building and running it yet, but I'm curious if it appears to be working with tests passing? Especially those related to the usage of `mknod` (_might work if that's run at runtime if build time isn't persisting `/dev`_).

AgustinRamiroDiaz:
> I haven't tried building and running it yet, but I'm curious if it appears to be working with tests passing?

@polarathene actually that's my main concern, tests do NOT pass at the moment of writing this, it'd be great if someone with Vagrant could check them to see if it's a problem from the Dockerfile or from the tests. The output is huge so I didn't want to paste it here, but can do it if you need it

polarathene:
@AgustinRamiroDiaz I don't get a good impression that it'll be easy to resolve that. It may be better to drop pursuing this further.

Are you aware of `eza`? It's a recent fork of `exa` where development is moving at a much faster pace and addressing many issues and PRs that have been open / stale for `exa`.

They've been [discussing the test suite as well](https://github.com/eza-community/eza/issues/146) with its issues, using Docker / Nix and migrating the test suite to something else that isn't as problematic for them.

ariasuni:
Closing this since exa is unmaintained (see #1243), you should take a look at [the active fork eza](https://github.com/eza-community/eza) (an alternative strategy has been used for tests).

PR #1233, opened 2023-08-11
Fix file extension for versioned shared libraries

### LINKED ISSUE #1105: Problem with penguin icon
As you may know, there is adorable little Tux icon assigned to .so files. But only .so files. I guess files with extensions so.*.* (i.e so.1.1) also counts as shared object files, but they lack little penguin icon. Only files with extension .so posses it.

gierens:
Revise File::ext to extract "so" extension from versioned shared libraries, so instead of returning "3" for "libtest.so.1.2.3" it returns "so".

The parsing only kicks in when the last extension is a numeric value, and then keeps on going back through the numbers until it finds the "so" or not. Only checking for numbers after the "so" should be sufficient for nearly all libs. I only found two exceptions in my `/usr/lib`, some `.so.owner` and a `.so.sign`. Trying to match those would require a much more general approach which doesn't seem to be worth it here.

Resolves #1105

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -132,9 +132,30 @@ impl<'dir> File<'dir> {
     fn ext(path: &Path) -> Option<String> {
         let name = path.file_name().map(|f| f.to_string_lossy().to_string())?;
 
-        name.rfind('.')
+        match name.rfind('.')
             .map(|p| name[p + 1 ..]
             .to_ascii_lowercase())
+        {
+            Some(ext) => {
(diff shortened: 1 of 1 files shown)


ariasuni:
Would you like re-open that PR but for eza?

gierens:
Ah, yeah right ... I'll probably do that ... but I'll have to think about it for a bit because this is just one tiny special case of so many. And actually when I wrote this PR I was already considering whether there is a way to handle multipart extensions in a general fashion.

Issue #1044, opened 2022-03-17
Add `--git=auto` feature


ariasuni:
It would be cool to have an option to only enable `--git` when a git repository is detected. Not sure how practical it is to implement, though.

Freed-Wu:
How about `--icons=(auto|always|never)`? Like `--git=(auto|always|never)`.

ariasuni:
Git is only enabled when a Git repository is found, so this is already implemented.

PR #1239, opened 2023-08-19
Add --no-git option

Aehmlo:
<details>

<summary>Some unnecessary exposition about my use case</summary>

While updating my `home-manager` configuration recently, I discovered that `exa` now has a program module available, with the option `programs.exa.git` to [alias `exa` to `exa --git`](https://github.com/nix-community/home-manager/blob/bdb5bcad01ff7332fdcf4b128211e81905113f84/modules/programs/exa.nix#L45). I happily enabled this option but quickly found myself in situations where I didn't want the git status to shown. While I could use something like `$(which exa) -l`, I far prefer the ergonomics of a `--no-git` option (and it plays more nicely with other tools to boot).

</details>

This PR adds the `--no-git` option. This option overrides `--git` in all cases and disables showing the Git status of files in long view (`-l`). This enables users of shell aliases and the like to turn off git status display in an *ad hoc* manner

Somewhat tangentially related: #1044.

(This is my first contribution to this codebase; I've done my best to make all the appropriate changes, but I'd appreciate any pointers if there's something I've missed. I took inspiration from the `--no-icons` option as prior art.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ These options are available when running with `--long` (`-l`):
 - **-@**, **--extended**: list each file’s extended attributes and sizes
 - **--changed**: use the changed timestamp field
 - **--git**: list each file’s Git status, if tracked or ignored
+- **--no-git**: don't list Git status (always overrides --git)
 - **--time-style**: how to format timestamps
 - **--no-permissions**: suppress the permissions field
 - **--octal-permissions**: list each file's permission in octal format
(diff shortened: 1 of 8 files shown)


cafkafk:
Hey, fyi, [exa is super dead](https://github.com/ogham/exa/issues/1139#issuecomment-1687228854), but if you'd like to, consider submitting this pr to [eza](https://github.com/eza-community/eza).

Aehmlo:
@cafkafk Thanks for the head-up. I just created eza-community/eza#168.

PR #1252, opened 2024-05-24
docs: Fix install command for Arch

AFAgarap:
`exa` is replaced by `eza` in Arch for some reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,9 +115,9 @@ On Alpine Linux, [enable community repository](https://wiki.alpinelinux.org/wiki
 
 ### Arch Linux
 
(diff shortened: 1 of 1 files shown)


phip1611:
Because the maintainer is not reachable :)


(This repository isn’t archived because the only person with the rights to do so is unreachable).
- https://github.com/ogham/exa/blob/master/README.md


The new home is https://github.com/eza-community/eza

ariasuni:
The first line in the README is already telling people to go see check eza, also it would be confusing to tell people to install eza in the exa repo…

Issue #1243, opened 2023-09-05
exa is unmaintained, please use the active fork exa instead


ariasuni:
exa is unmaintained, please use the [active fork eza](https://github.com/eza-community/eza) instead.

A lot of work has already been put into it by other people, and I’ll be porting my ongoing work to this fork. Sadly, @ogham seems unreachable so I can’t really archive the project… for now, I’ll redirect people to eza, from the README and issues page.

A page is turning for me: it was great maintaining exa on my own for a while but it wasn’t sustainable and I was in burnout, I’m so glad people finally created a fork that I could contribute to instead of having to handle the whole thing.

Aehmlo:
Second occurrence of `exa` in the current title should probably read `eza`?

kohane27:
Thank you so much for keeping us updated on the situation. I'm sure we all appreciate your effort to keep this project going! <3

ariasuni:
> Second occurrence of `exa` in the current title should probably read `eza`?

Thanks, fixed the typo x)

rugk:
You can [archive this repository](https://docs.github.com/github/creating-cloning-and-archiving-repositories/archiving-a-github-repository/archiving-repositories), to mark it as unmaintained. People then cannot submit PRs and there is a big warning that this is archived.

IMHO this would make it more obvious this is not maintained anymore.

Anyway, thanks for maintaining it so long! :hugs:

nikelborm:
@ariasuni just an idea: send unified instruction how to move PRs to new repo as comments of currently open PRs in this repo

ariasuni:
I’m not really sure how I would that myself. It would probably mean adding several upstream to the same repository, but I remember it being not very obvious.

Freed-Wu:
Hey, any one know what happen about @ogham? If some accident happened, maybe github official can give some maintainer a higher priority to transfer this repo after contace with github, which can keep all original issues and pulls.

ariasuni:
I honestly don’t know what happened to ogham, no activity since November 2021 and he didn’t answer to my e-mail from 2021-07-18.

YDX-2147483647:
FYI, ogham's website published a post on 2022-09-15, according to https://bsago.me/tech-notes/feed.json. ([sillysaurusx](https://news.ycombinator.com/user?id=sillysaurusx) found it 4 days ago at [Y Combinator](https://news.ycombinator.com/item?id=37416814)) Therefore I guess ogham is fine but too tired…

[...]

PR #1258, opened 2026-03-27
replace users with uzers

alexanderkjall:
Hello. users is unmaintained and affected by [RUSTSEC-2023-0059](https://rustsec.org/advisories/RUSTSEC-2023-0059.html).


ariasuni:
exa is unmaintained (see https://github.com/ogham/exa/issues/1243), and we already use uzers [the active fork eza](https://github.com/eza-community/eza).

